PR #4, opened 2014-11-06
Added bin/assume-unchanged to prevent modified config files appearing as...

pbinkley:
... modified in git status

I've added a shell script to run "git update-index --assume-unchanged" on a specified list of files, and I've added a section to README.md explaining it. New template files should be added to bin/assume-unchanged at the point when we want to freeze them.


pbinkley:
What a mess. Ignore it.

PR #27, opened 2015-01-14
Admin user's public profile

criedlberger:


pgwillia:
Looks like this introduces a regression in existing tests:

```
FFFFFFFFFF...

Failures:

  1) catalog searching p finds {P,S}
     Failure/Error: f.apply_depositor_metadata('dituser')
     NoMethodError:
       undefined method `group_list' for "dituser":String
     # ./app/models/concerns/hydranorth/model_methods.rb:11:in `apply_depositor_metadata'
     # ./spec/features/catalog_search_spec.rb:9:in `block (3 levels) in <top (required)>'
     # ./spec/features/catalog_search_spec.rb:6:in `tap'
     # ./spec/features/catalog_search_spec.rb:6:in `block (2 levels) in <top (required)>'
...
```

pgwillia:
I'm worried that this doesn't actually address the public profile of admin users (where dit.test@ualberta.ca is an admin user). As a public user (not logged in) I can see all of the admin's activity:
image
While it prevents future edits to uploaded files via the web gui -- because they don't appear here:
image Also the deposited files are never made public.
image
But they do appear in Solr and Fedora.
image

pbinkley:
I've merged from master, and the first few tests are failing. @criedlberger , would you mind having another look at this? There was a conflict in spec_helper.rb involving your Devise line, and maybe I'm not resolving it correctly. (Same applies to #59 )

weiweishi:
Tests passed except for #83

Issue #22, opened 2015-01-05
8.2 Link CC licenses to CC site


pbinkley:
as a user, I want CC licenses in the item display page to be linked to their descriptions in the CC site, so that I understand the license terms (note: Sufia pulls up a static description of the license - we could add the link there) (contributable?) size:small priority:medium

pbinkley:
This should be simpler once we finish #84 (do view tests with Selenium)

criedlberger:
Link CC licenses to CC site #59

PR #29, opened 2015-01-16
Licenses dropdown menu 

weiweishi:


pbinkley:
I notice two things that I'm not sure of here:

- _rights_modal.html.erb contains the "share-alike no-derivs" license, which we noted on Friday was self-contradictory (if you can't make derivatives, there's nothing to share-alike). Should this be updated? Or will this help file be overriden by @criedlberger 's changes in #22 ?
- sufia.rb contains a change in the FITS path in config.fits_path. Was this intended to be part of this pull request?

weiweishi:
I've removed the _right_modal.html.erb file from my commit since @criedlberger's change in #22 will override this file. 

FITS path was added to the sufia.rb file by the playbook and was not intended for this change. I've modify the file. 

The current head of this branch has already been merged with the master branch. Hope there won't be any further complication in merging the pull request, @pgwillia

pgwillia:
Sorry, still seeing issues with the added app/views/batch/_metadata.html.erb

PR #50, opened 2015-01-19
create a default admin user at application start resolves #36

### LINKED ISSUE #36: 9.7 Default admin user
As a developer or sysadmin, I want a default admin account to be part of the deployment package, so that I can save time by not having to create one

Done:
- [x] a default admin user with a dummy email address is part of the dev deployment
- [ ] a default admin user with a suitable real email address is part of the test and prod deployments
- [ ] stretch goal: in dev, allow developer to set the email address as part of their local environment (e..g use the email associated with their Github account?)

pgwillia:
Is modeled on DMP. Will need to change the password for test/production!

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,13 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+
+admin = User.new({
+      :email => "dittest@ualberta.ca",
+      :password => "password",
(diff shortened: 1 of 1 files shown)


pbinkley:
Would it take much work to enable it to use an email address stored in a local non-version-controlled file, like .vault_pass.txt? That way I could always have the local admin user send notifications to my own email address, rather than having to look for messages in dittest when I'm testing email stuff.

pbinkley:
After discussion, determined that the configurable email address is more trouble than it's worth.

PR #59, opened 2015-01-21
Link CC licenses to CC site

criedlberger:


pbinkley:
I've merged master into this branch to pick up recent changes, and the tests now fail. I'm not sure whether that's just because of my environment though. @criedlberger , would you mind having a look at the current state of this and see if the tests pass in your environment?

pgwillia:
I don't think this change ended up in the right path. Should _rights_modal.html.erb be in app/views/generic_file or app/views/generic_files? I'm thinking it should be the later based on where it is in sufia: [app/views/generic_files/_rights_modal.html.erb](https://github.com/projecthydra/sufia/blob/master/app/views/generic_files/_rights_modal.html.erb)

Issue #88, opened 2015-02-12
9.5.a No registered user can deposit into thesis collection


criedlberger:
See issue #34

pgwillia:
Proposed solution: have a thesis collection created on application installation owned by admin.

PR #98, opened 2015-02-23
No registered user can deposit into thesis collection

criedlberger:


pbinkley:
Note that this limits to the specific admin user dittest@ualberta.ca. When we move to deployment we'll need to think out what out-of-the-box admin user(s) there will be and how they are assigned to this collection.

pbinkley:
closes #88

PR #99, opened 2015-02-24
Selenium testing

weiweishi:


pbinkley:
@weiweishi , I'm curious about one thing: does properties.yml belong in the root directory, if it's only used in tests? Could it move down to spec-views, just to keep things clean?

And does this PR mean you're ready to come back from vacation?

PR #134, opened 2015-03-13
bump version of hydra-jetty

pgwillia:
To test this you'll need to replace your /var/www/sites/hydranorth/jetty dir:

```
rake jetty:stop # killall java
rake jetty:clean
rake sufia:jetty:config
rake jetty:start
```

This resets fedora and solr so all existing data would be lost.


pbinkley:
All tests pass.

PR #144, opened 2015-03-24
Nested collections fixes #75

### LINKED ISSUE #75: ERA_Collections: Nested collections
As admin I want to be able to create nested collections so that I can maintain a logical structure of broadly-focused collections with more sharply focused sub-collections
- [x] have link-to-parent and link-to-child metadata elements to enable up and down navigation
- [x] display links
- [ ] distinguish community and collection with a flag, so that the navigation can imitate current ERA with browsing of communities first, then links down to collection
- [ ] stretch goal: enable faceting by collection of collection, so as to allow searching across all nested collections

pgwillia:
* add requirements to collection model
  * cannot collect self
  * index collection it belongs to
* given a nested collection make sure you can view it
* add collections to collection in UI
 * side effect is giving 'files shared with me' batch options
* capybara feature test

### DIFF
--- a/app/helpers/batch_edits_helper.rb
+++ b/app/helpers/batch_edits_helper.rb
@@ -6,8 +6,6 @@ module BatchEditsHelper
   end
 
   def render_check_all
-    unless params[:controller].match("my/collections")
-      render partial: 'batch_edits/check_all'
-    end
+    render partial: 'batch_edits/check_all'
   end
 end
--- a/app/helpers/sufia/dashboard_helper_behavior.rb
+++ b/app/helpers/sufia/dashboard_helper_behavior.rb
@@ -40,6 +40,14 @@ module Sufia
       params[:controller].match(/^my\/files/)
     end
 
+    def on_my_collections?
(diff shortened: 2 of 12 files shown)


pbinkley:
It's letting me insert a collection to a child collection, and end up with Collection 3 includes Collection 2 and Collection 2 includes Collection3
collectionloop

pgwillia:
Isn't that the expected behaviour?

pbinkley:
I thought you fixed it to prevent loops - but no, I'm mixed up, that was just to prevent self-inclusion. Sorry about that.

pbinkley:
But: a UX problem: when you click "Add Collection", you get a menu of collections to add to the current one, but you don't have an indication of the current one now. It's easy to lose focus at the point and forget what the parent collection was. (And it will get worse when we have all the ERA collections in that list). Could we add a green-shaded message to the selections screen, like "Select collection to add to Collection X"?

pbinkley:
Decided to defer consideration of the usability issue until its priority is clear, given that this will be used mostly by admin if at all.

PR #154, opened 2015-03-27
minor modification to add missing citation field and dataset as a resource type

weiweishi:


config/initializers/sufia.rb
```diff
@@ -102,13 +104,9 @@
...
   # config.id_namespace = "sufia"
 
-  
-  # Specify the path to the minter-state file
-  config.minter_statefile = "tmp/minter-state"
-
   # Specify the path to the file characterization tool:
   # config.fits_path = "fits.sh"
-  config.fits_path = "fits"
+   config.fits_path = "/usr/local/bin/fits-0.6.2/fits.sh"
```
pbinkley: I think some unrelated changes got included here - minter-state and fits path should stay as they were, shouldn't they? I'm getting the old "Permission denied @ rb_sysopen - /tmp/minter-state" error.

config/locales/sufia.en.yml
```diff
@@ -211,8 +213,3 @@ en:
         description: "Usage information for abstract or summary"
         tag: "Usage information for keyword"
         rights: "Usage information for rights"
-
-  hydranorth:
```
pbinkley: And this section is needed by Chris's All Files change.

weiweishi: Sorry I thought the rebase would take care of it. Probably didn't do it
right. It's fixed now.

Issue #30, opened 2015-01-19
9.1 Metadata in forms


pbinkley:
As a user, I want to be able to describe my deposited item according to ERA norms to enable discovery and reuse
	- start with one item type (image?)
	- Metadata team will provide specs on needed changes with examples to existing Sufia forms
Note: Full-featured image deposit will depend on #57
Note: Dependent on Data Dictionary #103
Done: 
- [x] have forms with all of the fields (as spec'ed by metadata team) needed to describe the items
- [x] stored properly in Fedora, indexed and searchable properly in Solr, 
- [x] stretch goal: integrate with external ontology e.g. geo names

weiweishi:
I'm really interested in taking on this task. But it may require more than one pair of eyes from the dev team to get this working. Can we discuss the approach we'd like to take going forward on this issue?

sfarnel:
maybe we can touch base at the Wed standup?

On Tue, Jan 20, 2015 at 12:33 AM, Weiwei Shi <notifications@github.com>
wrote:

> I'm really interested in taking on this task. But it may require more than
> one pair of eyes from the dev team to get this working. Can we discuss the
> approach we'd like to take going forward on this issue?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ualbertalib/HydraNorth/issues/30#issuecomment-70614723>
> .
>



-- 
Sharon Farnel
Metadata Coordinator
University of Alberta Libraries
sharon.farnel@ualberta.ca
780-492-3685

weiweishi:
Some research on Authority Control in Sufia:
* questioning_authority: A gem used by the Hydra community to build local authority list, or a wrapper around existing authorities. Need to investigate more on how authority control information is handled in RDF and Fedora 4. 
* example to build drop down menu: https://github.com/curationexperts/mira/blob/5856e409d743fbc2e6fc026274972f5e10d4fb4e/app/helpers/contribute_helper.rb#L10
* example to get the long name (label) for the term (for example URI):
 https://github.com/curationexperts/mira/blob/276bdfe1ca3e2d465f6067a0dcc784bcac606efb/app/models/forms/capstone_project.rb#L18

weiweishi:
@sfarnel Sufia uses Schema.org microdata to align with Blacklight which uses schema.org for bibliographic records. This includes the item types. Could you please review if my following map is correct:
    "Book" => "http://schema.org/Book",
    "Book Chapter" => "http://schema.org/Book",
[...]

PR #156, opened 2015-03-30
Modified the Collection form to reflect the HN data dictionary

weiweishi:
as part of #30 

Also removed the duplicated methods in collection_controller_behavior.rb (those that were not changed and are extra duplicate of methods in Sufia and will be inherited from Sufia)


pgwillia:
This looks good.  ```rspec spec-views/collection_form_spec.rb``` pass but running ```rspec spec-views``` stalls but this is the same behaviour I get on master.  I'm comfortable merging :+1:

PR #143, opened 2015-03-24
New Generic File upload form

weiweishi:
#129 as part A of #30. 

Note the instruction of running selenium tests is in the note for #129

Issue #130, opened 2015-03-09
9.1 b Metadata form for special reports


weiweishi:
As a user, I want to be able to describe my deposited computing science/structural engineering reports according to ERA norms to enable discovery and reuse as part of #30 
Done:

- [ ] have the acceptable workflow to switch to special report deposit forms. 
- [ ] have forms with all of the fields (as spec'ed by metadata team) needed to describe the items
- [ ] have the deposited items added to the proper collection
- [ ] stored properly in Fedora, indexed and searchable properly in Solr,

PR #163, opened 2015-04-01
Special reports form for CS Technical Report and Structural Engineering Report

weiweishi:
As part of #30 and #130. Related pull request #143, #156. 
- Added to batch_controller_behavior and file_controller_behavior to handle special report ids. 
  - deposited items will be added to the corresponding collections. It assumes the collections have already created in the system. Please add the collection ids to the config/initializer/sufia.rb. 
- Added presenters and forms for special reports
- Added selection of special reports on the upload form


pbinkley:
When I upload a Structural Engineering Report, I get an error `undefined method cstr_collection_id` at `app/controllers/concerns/hydranorth/batch_controller_behavior.rb:27`, which has 
`@collection = Collection.find(Sufia.config.cstr_collection_id)`. In `config/initializers/sufia.rb` it looks like you've commented this string out:

```  # config.cstr_collection_id = "6q182p35f"
  # config.ser_collection_id = "6q182p36q"```

How will we manage this in production? Will it have to be based on the final NOID for the collection? Is there any way to base it on a property of the collection rather than hard-coding the id in the config? @criedlberger is looking at a similar question for the non-editable thesis problem: can we set a property on a collection to govern whether a user can edit their own items in that collection. Maybe there's an overlap in methods?

weiweishi:
It is not difficult to implement such method for the collection model. I
actually had a method to search based on resource type but thought it would
be expensive to traverse all collections and search for the right one each
time user make a deposit. My decision was based on the assumption that the
NOID will not change in production and the list of values can be managed by
Ansible to add to the initializer config file. But I can definitely put it
back in.

For Chris' problem, we can certainly set additional property, possibly
within the permission hash. But I don't know enough about Cancan to see
this through.

pbinkley:
Ok, on that basis, could we add the appropriate CSTR and SER collections to the dev environment in seeds.rb, the way we do with the Thesis collection? All of them will have to be replace in production, but this way we won't get errors when trying things out in dev.

weiweishi:
Okay.  seed.rb is to seed the database. I will create a rake task to
initiate the seed objects and add the step to the playbook.

Weiwei

pgwillia:
Don't need to add a rake task or step to playbook.  ```rake db:seed``` already exists and is called by ```rake db:setup``` in the playbook.

pbinkley:
We should look at handling the Thesis collection in the same way for consistency, since it's currently in `db/seeds.rb`.

weiweishi:
Now handling the initiation of these collection in seed.rb. 
Also added/changed the creation of existing data to avoid duplication if rerun rake db:seed. 

Please restart httpd after rake db:seed, as it writes the collection ids to config/initializers/sufia.rb, and will need a restart to make the modification in effect.

pbinkley:
I've rebased on master, and the tests pass, but I'm still getting an error when I try to create an SER item manually: when I submit the metadata form for a new item from the SER form, I get this error. 
error

PR #172, opened 2015-04-08
configuration and dependencies for clamav

pgwillia:
Resolves ```Virus checking disabled, /tmp/RackMultipart20150402-724-1fwm0sz not checked in logs```

* need to look at travis exclusion?
* requires (yum install) clamav-devel for bundle install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ before_install:
   - export PATH="$PATH:$(pwd)/fits"
 services:
   - redis-server
+before_install:
+  - sudo apt-get update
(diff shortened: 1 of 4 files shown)


pbinkley:
I've added installation of the clamav dependencies to the .travis file, and it now runs on Travis. Clamav installation is working on Travis (though the unrelated collections test is still failing). Ultimately we'll want to test how virus-ridden uploads are handled and reported, so this belongs in the dev environment, I think. @pgwillia , if this looks good to you, would you merge?

pgwillia:
OWASP V16.3

PR #174, opened 2015-04-09
Theses can be edited by admin, but belong to user

criedlberger:


pbinkley:
A bug: when I share a file from admin to a regular user, that user does not have an edit link when they view the item, which is good, but they still get an edit link in the "Select an action" dropdown on the My Files list (see screenshot)
editbug

pbinkley:
Sorry - one last problem - There are two conflicts when I rebase on master (which now includes the sufia6 upgrade). Could you resolve those and recommit?

criedlberger:
Close and replace with #199

PR #199, opened 2015-04-13
Theses can be edited by admin, but belong to user

criedlberger:


spec/controllers/my/files_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        expect(assigns[:document_list].map(&:id)).to include(@my_file.id)
+        # doesn't show files shared with other users
+        expect(assigns[:document_list].map(&:id)).to_not include(@unshared_file.id)
+      end
+  end
+
+  describe 'User can edit and not edit' do
+    before do
+      visit "/dashboard/files"
+
```
pgwillia: This before doesn't have a closing end so the following tests aren't executed (should be 4 examples, only registers two).

spec/controllers/my/files_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        expect(page).to have_content( "Edit File" )
+      end
+
+      it 'user cannot edit' do
+        click_link "My Files"
+        within("#document_#{@read_shared_with_me.id}") do
+          click_button("Select an action")
+        end
+        expect(page).to_not have_content( "Edit File" )
+      end
```
pgwillia: These are good feature styled tests (click_link, click_button, etc) which test the view rather than just the controller so should be in spec/features not spec/controllers.

spec/controllers/my/all_controller_spec.rb
```diff
@@ -43,34 +43,6 @@
...
-        expect(assigns[:document_list].map(&:id)).to_not include(@my_collection.id)
-        # doesn't show shared files
-        expect(assigns[:document_list].map(&:id)).to include(@shared_file.id)
-        # doesn't show other users' files
-        expect(assigns[:document_list].map(&:id)).to_not include(@unrelated_file.id)
-        # doesn't show non-generic files
-        expect(assigns[:document_list].map(&:id)).to_not include(@wrong_type.id)
-      end
-    end
-  end
```
pgwillia: First thought is why did you remove the old test? 

Second thought is don't we want to verify that non-admin users get an unsuccessful response?

spec/controllers/my/files_controller_spec.rb
pgwillia: Is there a test for the I Can Edit / Depositor: me case?

spec/controllers/my/files_controller_spec.rb
criedlberger: Corrected

spec/controllers/my/files_controller_spec.rb
criedlberger: I have added the test.

spec/controllers/my/all_controller_spec.rb
criedlberger: Non-admin users should not be seeing the All Files tab at all.

pgwillia:
Looks good.  Waiting on travis.

Issue #168, opened 2015-04-02
Different resource type list for regular user and admin


weiweishi:
As a regular user, I would like to deposit items with the allowed resource type, and create personal collection without the options of selecting resource type.  
As an admin user, I would like to see additional resource types such as Computing Science Technical Report, Structural Engineering Report and Thesis. I would like to create official collections with these resource_types, as well as to see thesis as a valid resource type during migration. 
- [ ] resource type list on batch upload form, record editing form stays the same for regular user, but have additional resource types for admin user
- [ ] admin user can create collections with specific resource types while regular users don't get to choose resource type(s) when they create collections.

pbinkley:
Closed by #200

PR #200, opened 2015-04-13
Admin resource type list

weiweishi:
...for regular user and admin

This is to address #168 and should also solve the potential resource type conflict issue (including thesis etc) during ERA migration.


pgwillia:
Should there be a test for the new functionality?

weiweishi:
I've created feature tests for the PR. Please merge it whenever someone has a chance. 
Weiwei

Issue #201, opened 2015-04-14
RSOLR error when opening collection detail page


weiweishi:
Now there were about 130 items in the collection, it shows the following error message when opening the collection description page:

It seems that the query includes all the member ids within the collection, which makes the request too big to process. It's a potential bug upstream. 

(can't do a screen capture right now but will do one later). 

RSolr::Error::Http in Collections#show
Showing /var/www/sites/hydranorth/vendor/ruby/2.1.0/gems/sufia-6.0.0/app/views/collections/_show_descriptions.html.erb where line #3 raised:

RSolr::Error::Http - 413 Request Entity Too Large
Error: Error: Request Entity Too Large: head

[...]

Issue #157, opened 2015-03-30
Migrate multi-file objects


pbinkley:
As zips or collections.
must validate checksums for individual files
instructions to users.

weiweishi:
Script will create a zip file if multiple DS datastreams is available in the foxml, and upload the zip file - date_uploaded is the upload date for the last DS datastream.

Issue #150, opened 2015-03-27
migration - users


pgwillia:
Bring over users first so that items can be linked to them as depositor.

* ensure devise unconfirmed and can gain confirmation of email instructions
* create field for ERA user id
* some users may have CCID associated with their account - can we discard this? And deal with it later when Shibboleth is part of our environment? (Question for Amanda)

pgwillia:
Will admin users be created as part of the migration?  This would solve #183

weiweishi:
current admin users in ERA will be migrated. New admin users won't be there
- e.g dittest etc.

pbinkley:
That's one where we'll want to force a password change, or at least change
the password ourselves pre-migration. The dev team should not need access
to the real eraadmin password to work on the migrated data.

Peter


Peter Binkley
Digital Initiatives Technology Librarian
Information Technology Services
peter.binkley@ualberta.ca

2-10K Cameron Library
University of Alberta
Edmonton, Alberta
Canada T6G 2J8

phone 780-492-3743
fax 780-492-9243

weiweishi:
We can reset the password on Production manually for the admin users after the migration. Does #183 only refer to real eraadmin?

pbinkley:
No, eraadmin does not need to change.

Issue #149, opened 2015-03-27
migration - archival foxml datastream


pgwillia:

weiweishi:
Added archival foxml (downloaded from era public interface) to generic file.

PR #233, opened 2015-04-30
Era migration

weiweishi:
In this PR, it has:
* migration:eraitem which migrates each item from the metadata directory (argument from the rake task)
use: rake migration:eraitem['lib/tasks/migration/test-metadata/metadata']
* migration:era_collection_community which migrates collections and communities in the metadata directory
use: rake migration:era_collection_community['lib/tasks/migration/test-metadata/collection']
* migration:user_migration which migrates users in the user files (data cleaned from ERA mysql tables user and author) 
use: rake migration:user_migration['lib/tasks/migration/users/users.txt']
  ** note: this has only been tested with small dataset - need to proceed with cautious due to the potential of sending emails to migrated users
  ** TO-DO: migrate user avator/profile images 
* modified users table and user model to handle legacy password and other information migrated from ERA
*  modified collection model to better handle rsolr requests (that has caused #201)
This PR should address #149 #148 #150 #157


pbinkley:
Replaced by rebase-test branch, so a new pull request has to be started.

PR #234, opened 2015-04-30
clamav 'no supported database files found' failing travis builds for some workers

pgwillia:
Not sure if this is the required fix.  
Suggested by https://github.com/psu-stewardship/scholarsphere/commit/3bd94a9233abe225de081cf8f0580e78f4238253

[clamav](https://wiki.archlinux.org/index.php/ClamAV#Error:_No_supported_database_files_found) suggests ```freshclam -v```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ services:
   - redis-server
 before_install:
   - sudo apt-get update
-  - sudo apt-get install libclamav6 libclamav-dev clamav clamav-data
+  - sudo apt-get install libclamav-dev clamav-freshclam
 before_script:
   - redis-cli info
   - bundle exec rake jetty:start


pbinkley:
When I looked into this when I was first setting up Travis, it appeared that the freshclam update was no longer needed if you install clamav-data: https://github.com/ualbertalib/HydraNorth/commit/dae8b1bef41b6856a52235a6126954581f86c877 , based on https://github.com/duke-libraries/dul-hydra/issues/1193 (which is 6 months later than the PSU issue you referenced) . So I don't think this is what's causing the intermittent failures. Maybe we just need a sleep of a few seconds to make sure clamav-data has finished installing?

pbinkley:
Or just do the freshclam update as a belt-and-suspenders step, I guess.

pbinkley:
It made a test pass that failed last night because of clamav, so at least it isn't breaking anything. Here's hoping it's the solution.

Issue #182, opened 2015-04-10
Thumbnails not generated for multimedia files


pbinkley:
PDFs, possibly others
on production: tottenham.library.ualberta.ca

pgwillia:
From the [fixtures](https://github.com/ualbertalib/HydraNorth/tree/seed/spec/fixtures) that come with Sufia the following create broken thumbnails
* sample_mpeg4.mp4 
* countdown.avi

and 1.5mb-avatar.jpg would not upload.

pbinkley:
So: we need to figure out why the tools that resque is using to create thumbnails from multimedia files are failing without causing the default thumbnail to be used.

criedlberger:
In order for video and audio files to have proper thumbnails the following line must be uncommented in config/initializers/sufia.rb.

config.enable_ffmpeg = true

criedlberger:
Closed with pull request #261

PR #261, opened 2015-05-12
Thumbnails not generated for multimedia files #182

criedlberger:


pbinkley:
@criedlberger , The new tests are failing on Travis with "No such file or directory @ rb_sysopen - /home/travis/build/ualbertalib/HydraNorth/spec/fixtures/countdown.avi". Could you commit this file? Hopefully it's not too big...

config/initializers/sufia.rb
```diff
@@ -144,6 +144,7 @@
 
   # Specify the path to the file characterization tool:
   # config.fits_path = "fits.sh"
+   config.fits_path = "fits"
   config.fits_path = "fits"
```
pbinkley: @criedlberger Committing configuration changes here - I'm not sure whether we still need to keep this path as something that gets set on installation, but as long as it is we shouldn't add multiples.

criedlberger:
I have added missing file

PR #291, opened 2015-05-22
added new spec test for handling special report ids in batch upload job

weiweishi:
This is to add additional spec tests for PR #197 for the batch_update_job, which handles the new deposit from the deposit form.


pgwillia:
This looks like it was already included in https://github.com/ualbertalib/HydraNorth/commit/7db2909cf0a97704f2bbfae635f982a8730c2472

PR #326, opened 2015-05-24
Migration fixes

weiweishi:
* exclude objects with license files or long (longer than 250 characters) license text from migration, until we have established a plan on how to deal these objects. 
* close mysql connection after each user's migration as ActiveRecord doesn't close the connection automatically.


pgwillia:
I'm going to merge this but the sample items need to be re-thought because the majority of them are not migrated because of licenses.

pgwillia:
The 'license is a file or text is longer than 250 characters' text appears for every item.

weiweishi:
Hmm. Okay. I will get more test metadata - After Henry and I walked through
the migration tasks, several areas of change are identified. I will be
working on updating this PR. Thanks!

PR #329, opened 2015-05-25
test for relevancy option for sorting resolves #323

### LINKED ISSUE #323: Browse results: sort by "New items first" when relevance is irrelevant
In single-value browse queries (e.g. items in this collection or items with a given publication date), sorting by relevance makes no sense, since every item has the same score. In these cases, the default sort should be "New items first" (i.e. sort by date_uploaded, descending).

pgwillia:


pgwillia:
Missed some dependencies.  Working this out.

pgwillia:
Looking for feedback.  Changed from misguided view test to feature test.

Issue #295, opened 2015-05-22
"Select an action" should not show on collection pages without logged in user


weiweishi:
the "select the actions " dropdown menu still include edit/remove etc although user require permission to do the actual change, it's confusing and should be removed from the list.

PR #334, opened 2015-05-25
"Select an action" should not show on collection pages without logged in user #295

criedlberger:


weiweishi:
When user is not logged in (or logged out) they will see an error message as such:
http://plano.library.ualberta.ca/collections/bv73c0908

image

PR #335, opened 2015-05-25
Browse functionality.

nnunn:


weiweishi:
Seeing the following error message after checking out the browseux branch - have restarted with bin/restart_all. 
image

nnunn:
Did restart all work? I have a feeling this is because there is more
records than I am working with. I will dig into it.

nnunn:
Ok that bug is now fixed.

weiweishi:
Looks good @nnunn. One thing though:
For the Item Type, only the first item type shows in the list, is it possible to make it populate with more item types? example: http://plano.library.ualberta.ca/browse
image

nnunn:
ahh sorry - I was using less results so I could test it...

That fix is committed and pushed now.

nnunn:
wait it's not up yet. 2 minutes.

nnunn:
Ok that fix is up now.

weiweishi:
Thank you!

PR #338, opened 2015-05-26
Special report detail page

### LINKED ISSUE #296: special report ids should be showing on the record page.

weiweishi:
PR includes:
* make the order of fields on special reports record page to be consistent with regular generic file
* add a rake task to update the resource type for special reports (they are currently set as 'report' during migration) - Metadata team is going to evaluate the metadata value they'd like to use, some functions may require refactoring/new development after their decision. This rake task will fix #296

### DIFF
--- a/app/presenters/hydranorth/cstr_presenter.rb
+++ b/app/presenters/hydranorth/cstr_presenter.rb
@@ -3,7 +3,7 @@ module Hydranorth
     include Hydra::Presenter
     self.model_class = ::GenericFile
     # Terms is the list of fields displayed by app/views/generic_files/_show_descriptions.html.erb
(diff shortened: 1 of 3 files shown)


pgwillia:
What should this look like when it's working?  Did a bin/reset-all and ran the migration rake tasks from the README.  Type of Item is "Report"
image
image

weiweishi:
It should look like this:
[image: Inline image 1]
In order for the rake task to work, the collection has to be migrated
already and existed in the system. Then run
rake hydranorth:update_special_itemtype
It will update the resource type "report" to "computing science technical
report" if this item is a member of "technical report".

pgwillia:
image

PR #353, opened 2015-05-27
updating numbers on graphs on front page

nnunn:


nnunn:
Not sure if I can squeeze this last minute change in, but here it is if I can!

PR #359, opened 2015-05-28
fix for characterize rake task to get the complete list of objects fo…

weiweishi:
…r the queue, instead of only the first 10.


pgwillia:
Is the characterize rake task documented anywhere?

weiweishi:
I will update the README.md as part of the PR.

pgwillia:
This task doesn't get any logging that I can see. Does it need any?

PR #455, opened 2015-07-06
dataverse object migration, with changes to the data model

### LINKED ISSUE #67: Dataverse to HydraNorth
As a data librarian, I want to be able to ingest item-level metadata from Dataverse into the DAMS, so that Dataverse items are discoverable via the HydraNorth discovery layer while they are still accessed in Dataverse.

Part of epic #69

pgwillia:
work to date fixes #67


pgwillia:
@pbinkley @weiweishi I made some changes so that all of the tests are passing.  I'm not 100% sure that changing the sample data for community and collection to respond to the change in "xmlns:dcterms"=>"http://purl.org/dcterms/" is the correct thing to do. Why was this change introduced?

johnhuck:
when I check the DCMI page, I find that the correct namespace is in fact http://purl.org/dc/terms. I'm not sure if I am responsible for introducing the http://purl.org/dcterms namespace into the code, but that is the namespace I was using in the records, which I now see was in error. I will make the necessary changes to my stylesheet for future metadata processing. The namespace in the code should revert to http://purl.org/dc/terms.

pbinkley:
Good that we caught that. I'll have time on the bus to revert Tricia's last commit with the changed namespace, fix the namespace in the test metadata in export_oai_dc_dcterms.xml, rebase on master and hopefully merge the PR.

PR #536, opened 2015-07-24
Layout and language of upload form - fixes #504

### LINKED ISSUE #504: Layout and language of upload form: Priority 2
As a depositor, I want to understand if the options presented on the deposit form are relevant to my situation, so I can make the right decisions about uploading my items.

Participants found the "Uploading special reports?" language on the upload form confusing - weren't sure if this language applied to them.

Participant who opened the DAMS agreement got lost in orphaned new window.

Possible solutions:
- [x] reposition the Special reports options below the DAMS agreement, or provide other ways to reduce visibility when not necessary
- [x] keep DAMS agreement visible ( keep on page with expandable text / or open new window with "I agree" box on that page / or open in same page with easy option to return using back button or breadcrumbs)

pbinkley:
- rearrange upload form elements
- include agreement text in upload form as expandable div
- update agreement text and move it into partial


pgwillia:
Clicking on the 
image
Doesn't prompt for Computing Science Technical Report:
image
This is what would be expected:
image


I see that we don't have a test for this.

pbinkley:
Hmm; I thought that was  preexisting bug, but I see it works on hydranorth
production. I'll try to get that fixed and tested this afternoon.

pbinkley:
I've fixed the CSTR problem and added a test. The test is a little clumsy (uses sleep(15) to wait for new page to load), since Sufia's javascript seems to evade Capybara's usual smartness about waiting for a new page with e.g. page.find("form#new_generic_file") (which should block until a form with that id appears).

pgwillia:
I'm still seeing the CSTR problem and your test is failing.

PR #580, opened 2015-09-02
updated the method to check for duplicated records during migration

weiweishi:
updated duplicated? method in migration to properly retrieve the duplicated records.


pbinkley:
@weiweishi , if you're ok with the way I've added a log message, I'll merge this.

weiweishi:
It looks good @pbinkley.

PR #594, opened 2015-09-03
Saml 1

mbarnett:
Adds support for SAML scenario 1 and includes some refactoring of the SAML authentication code.


app/controllers/users_controller.rb
```diff
@@ -20,19 +22,33 @@ def unlock_access
...
   def set_saml
-    if (params[:has_account] == 'no') || params[:user][:email].nil?
-      @user.ccid = @user.email
-      flash[:notice] = I18n.t('devise.omniauth_callbacks.success', :kind => 'Shibboleth')
-      sign_in_and_redirect @user, :event => :authentication
+    if params[:has_account] == 'yes'
+      existing_user = User.find_by_email(params[:user][:email]) if params[:user][:email].present?
+
+      # the user may have typoed the email they tried to link
+      # send them back to try again
```
pgwillia: One of the commitments the dev team made to our security expert @raywchan was to pass the [opportunistic OWASP checklist](https://www.owasp.org/images/5/58/OWASP_ASVS_Version_2.pdf). V2.18 asks us to not be this nice.

app/models/user.rb
```diff
@@ -17,19 +15,17 @@ class User < ActiveRecord::Base
...
-      return user
-    end
-    where(provider: auth.provider, email: auth.uid).first_or_create do |user|
-      user.skip_confirmation!
-      user.email = auth.uid
-      user.password = Devise.friendly_token[0,20]
-    end
-
+  scope :from_omniauth, ->(auth){ where('(email = ? OR ccid = ?) AND provider = ?', 
+                                         auth.uid, auth.uid, auth.provider).limit(1) }
```
pgwillia: This scope and ```, ->(auth)``` is new to me.  How does this work?

pgwillia:
Probably don't want to add changes to config/environments/development.rb

app/controllers/users_controller.rb
mbarnett: Makes sense. In order to make it act the same in the 'found actual user' and 'you literally typed gibberish' states, we will need to delete the CCID user regardless, and recreate it the next time they try to log in from shibboleth.

mbarnett:
Those changes seem to be automated by the vagrant? I guess it's worth asking why the vagrant is making changes that everyone is sharing that aren't reflected in source control?

app/models/user.rb
mbarnett: [Rails scopes](http://guides.rubyonrails.org/active_record_querying.html#scopes) are the preferred way of centralizing query logic in Rails models. They have a lot of nice properties, like composability and extensibility with all other user-written and built in Rails scopes, and many gems, that you otherwise have to do extra work to get if you use class methods to implement query functionality. So it's generally preferably to use scopes for containing query logic whenever possible.

->() is the Ruby 1.9 "stabby lambda" lambda syntax. In this case, we're declaring the scope to be implemented by a lambda that takes one argument, auth.

pgwillia:
I'm not sure if I have a *good* answer to this. My thinking is to provide
good default values in the github repo for anyone that may want to try this
out. The ansible playbook knows some things about the deployment
environment like default_url (configured in httpd) that are specific to the
application when deployed that way. It has similar changes for test and
production servers. These values are going to be overwritten every time the
playbook is run so if they are going to change there needs to be follow up
changes to the ansible playbook for HydraNorth. I don't think we've settled
the boundary between application configuration and system configuration
yet, right @redlibrarian?

redlibrarian:
IMO we haven't settled that boundary, and what we've seen (at least on the Blacklight project) is that things that originally lived in the application (and GitHub) have gradually moved over to the playbook (and Mercurial) in order to facilitate the way the sysadmins work, and the way they've set up our systems. I don't think we've ever taken a step back to figure out if this was the best way to do things - for many reasons, some of the them time constraints. At some point I would like to think about the best way to make this distinction.

mbarnett:
Right, right. Hard to remember this is a public project. 

So the typical solution I've seen to the problem of having environmental config files that are going to have different values for contributors is to explicitly ignore the file in git, rather than to rely on everyone keeping a bunch of uncommitted changes on their machine. Example configuration is typically provided by instead checking in an .example version of the file, and maybe adding a rake task to set things up by copying the example files over to their real equivalents.

On the subject of .gitignore, it's probably worth discussing ignoring the Gemfile.lock and db/schema.rb as well. Those are typically not committed and essentially act as nothing more than an endless source of conflicts for no meaningful gain.

weiweishi:
*I think* part of the reason that we decide to keep Gemfile.lock in the
repo is to allow system admins to create a RPM that uses the exact same
versions of gems as on dev machine to avoid any unpleasant surprise. This
was also a decision made earlier when we first started working on Rails'
projects. We never had the opportunity to revisit and probably worth a
discussion between dev and ops again.

*Weiwei Shi*

2-10L Cameron Library
Phone:(780)492-7802
Fax: (780)248-1209
Email: weiwei.shi@ualberta.ca

pgwillia:
Re: Gemfile.lock

November 2014 we decided to remove Gemfile.lock after a discussion.  I'd
already been through the cycle of in
<https://github.com/ualbertalib/HydraNorth/commit/f5361d72ef7c200b3cecc04adbb8fc12830b26f6>
and
out
<https://github.com/ualbertalib/HydraNorth/commit/b234875f87a0c91b6481052b1784d295953e1be4>
once
before.

March 2015 we decided to put it back: We discussed at the
retrospective/kickoff that it's valuable to have a point of reference for
which combination of gems are known to work together.

mbarnett:
The Gemfile.lock is actually platform dependent so it's something that really *shouldn't* be heading up to a server straight from a dev's machine. ie) if the last developer to commit the Gemfile.lock did so on a, say, Windows Machine, this can break on a Linux server, depending on Bundler version, gems involved, phases of the moon, etc.

Might be worth investigating generating it on a CI server, or locking down versions more tightly in the Gemfile to avoid surprises.

redlibrarian:
We've talked about having a big debrief (hopefully 1/2 or full day) to talk about what we've learned in both Blackligh and Hydra projects, as well as all the supporting technologies (ansible, vagrant, etc), but we haven't gotten to the point where that makes sense. I would like to revisit a lot of this stuff, in the light of +1 year's solid work on both projects - what was a good choice and what wasn't, in hindsight. I realize this may be extending the scope of this particular PR, but it's good to capture right now I think.

pbinkley:
With everyone developing on the same CentOS platform as deployed by
Vagrant, that should never happen. But yeah.

redlibrarian:
We had done a lot of work initially on having everyone work on the same CentOS image. In practice, on the Blacklight project, that didn't happen - Natasha developed on Mac and I developed on Debian.... getting tighter on our gem versions in the Gemfile is a good idea no matter what, IMO.

mbarnett:
Thanks all for filling me in on the motivations behind these decisions, y'all. Takes a mindset adjustment for me coming out of 6 years of doing Rails in private.

If anyone else wants an easy way to ignore changes to some of these files in the interim, I'm using

````
git update-index --assume-unchanged config/environments/development.rb
````

mbarnett:
This should now be ready to merge if you're happy with it, @pgwillia

PR #614, opened 2015-09-16
Remove clamav to improve performance

pbinkley:
This uses the same Travis config as the ffi-clamav project does. It saves about five minutes (23 instead of 28) and runs all our tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ services:
   - redis-server
 before_install:
   - sudo apt-get update
-  - sudo apt-get install libclamav-dev clamav-freshclam
-  - sudo freshclam
+# clamav installation copied from https://github.com/postmodern/ffi-clamav/blob/master/.travis.yml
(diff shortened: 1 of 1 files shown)


pgwillia:
Does the line
```
 - gem install ffi rubygems-tasks rspec yard
```
apply to us? My guess is not so I've removed it.

pbinkley:
I had errors without it - I think the ffi is necessary. But we'll see when travis finishes testing this commit.

pgwillia:
Our Gemfile.lock already has ffi as a dependency so should get installed. https://github.com/ualbertalib/HydraNorth/blob/master/Gemfile.lock#L207

pbinkley:
But not in time for ClamAV to be happy, was my impression.

pbinkley:
Well, it looks good - I must have misinterpreted an error my testing.
Thanks for catching this!

PR #616, opened 2015-09-17
Use Dataverse DOI link instead of download link fixes #494

### LINKED ISSUE #494: Dataverse to HydraNorth: download link
- [ ] Download link takes user to a dead end. Can the Dataverse DOI URI be substituted? Otherwise, can the link be deactivated?
Note: DOI is stored in identifier field. We would only want it to be used as the download link for items in the Dataverse collection.

pgwillia:
* refactor linkable so that helper and model can take advantage
* refactor search method in spec to helper so that it can be used wider

### DIFF
--- a/app/models/concerns/hydranorth/generic_file/doi.rb
+++ b/app/models/concerns/hydranorth/generic_file/doi.rb
@@ -1,7 +1,7 @@
 module Hydranorth
   module GenericFile
     module DOI 
-      include Linkable
+      include ::LinkUtils
 
       def doi_url
         self.identifier.first if ((!self.identifier.first.nil?) && (linkable? self.identifier.first))
--- a/app/models/concerns/hydranorth/generic_file/doi.rb
+++ b/app/models/concerns/hydranorth/generic_file/doi.rb
@@ -1,10 +1,7 @@
 module Hydranorth
   module GenericFile
     module DOI 
-      extend ActiveSupport::Concern
       include Linkable
-      included do
-      end
 
       def doi_url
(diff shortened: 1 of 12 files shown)


app/models/concerns/hydranorth/generic_file/doi.rb
```diff
@@ -0,0 +1,15 @@
+module Hydranorth
+  module GenericFile
+    module DOI 
+      extend ActiveSupport::Concern
+      include Linkable
+      included do
```
mbarnett: I don't think we need to define an included block if it's empty?, for the same reason, it's probably unnecessary to extend ActiveSupport::Concern if we're not taking advantage of it.

lib/linkable.rb
```diff
@@ -0,0 +1,8 @@
+module Linkable
```
mbarnett: Really, really mild nit, but I'm not sure about organizing a utility method like this? 

Naming wise, if I see 'include Linkable' in a class, I think that should be adding functionality to make that class linkable (by analogy with Hydra's Embargoable class, for instance), rather than just pulling in a utility function with that name.

It might be better to put this function in a file in lib/ (link_utils.rb or something), and then autoload it in application.rb:

```ruby
config.autoload_paths += Dir["#{config.root}/lib/**/"]
```

 This would also eliminate the need to require it in the application_helper.

pgwillia: Yeah ruby/rails organization still baffles me, not always sure where to put things. Just tried this and couldn't make the autoload part work.

pgwillia:
@mbarnett, I think I've addressed your concerns.  Thanks for your feedback!

mbarnett:
:+1:

PR #617, opened 2015-09-17
Ux

nnunn:


pgwillia:
Working through broken tests: currently 64 failures. A lot of them seem to be a change to the "Log in"/"Login" button. I'll work to resolve these.

nnunn:
Thanks - let me know if there is anything I can do.

pgwillia:
The problem I'm having is that we're using Sufia's [session_helpers.rb](https://github.com/projecthydra/sufia/blob/master/spec/support/features/session_helpers.rb) but the helper expects the button to say 'Log in' while the ux branch has changed this to  'Login' so the tests using that helper all fail.  You'd think there would be an easy way to fix this but the helpers seem to be all tied together in knots.  Suggestions?

mbarnett:
Best option is probably to either redefine Sufia::Features::SessionHelpers in our SpecHelper, or just remove this import entire https://github.com/ualbertalib/HydraNorth/blob/master/spec/spec_helper.rb#L14

and write our own Hydranorth::SessionHelpers and import that instead.

Either way you're basically just duplicating that method but replacing 'Log in' with 'Login' (or even better, stick the string in a locale file and use that instead of hard-coding it).

nnunn:
Or I can just change it back to Log in.

pgwillia:
I've worked past the Login problem.  No worries.  Now dealing with the changes to Collections from the dashboard affecting tests.

nnunn:
Seems Log in is actually the verb version of the word(s) so that is "more" correct. Sorry! I will make the change on the ux branch and commit.

tash

pgwillia:
I introduced sign_in to the locales and used it instead: https://github.com/ualbertalib/HydraNorth/commit/a7296a439e2848be79f4719cb87ba434627b47f1

pgwillia:
I'm seeing the reload bug on the "My Profile" page.  @nnunn is there an easy fix for that?
image

nnunn:
No - I'll need to take a look at it. I thought I had it fixed.

mbarnett:
This test is true on the profile page: https://github.com/ualbertalib/HydraNorth/blob/ux/app/assets/javascripts/custom.js#L122

because the icon is used for the joined date. So it just keeps reloading the page

nnunn:
ah ok I can wrap the files page time icon in a named div and only check that one. Should I make that change on the ux branch?

pgwillia:
Yes please!

pgwillia:
Maybe it's just 'cause I have my browser half size all the time that I notice these things, but...
image

nnunn:
yup I know about this issue. It's part of the responsive work I need to do next week.

nnunn:
Ok I've fixed the reload issue and pushed it up to the ux branch

nnunn:
Hey Tricia - is there anything else you need me to do to get this pull request done?

pgwillia:
Almost done.  Was weird thing with commits repeating 4 times.  I've squashed and am about to push, then will merge pull request.

pgwillia:
Uh oh.  Looks like inadvertent gem upgrade is causing new failures in tests :(

 Encoding::CompatibilityError:
       incompatible encoding regexp match (UTF-8 regexp with ASCII-8BIT string)

nnunn:
Is that something I can fix?

pgwillia:
I see two paths:
 1) pin the Gem so that the change doesn't impact us
 2) add ```# encoding: utf-8``` to all the places this is failing.

pgwillia:
This is our problem too: https://github.com/ruby-rdf/rdf-turtle/issues/13.  Will pin that gem.

nnunn:
Ok great

PR #636, opened 2015-09-29
fix download_path wrong number of arguments bug introduced by #616

pgwillia:


app/helpers/generic_file_helper.rb
```diff
@@ -21,8 +21,9 @@ def render_download_link text = nil
     link_to (text || "Download"), download_path(@generic_file), { id: "file_download", target: "_new", data: { label: @generic_file.id } }
   end
 
-  def download_path gf
-    GenericFile.find(gf.id).doi_url || sufia.download_path(gf.id)
+  # sufia.download path is from ActionDispatch::Routing::PolymorphicRoutes#polymorphic_path
+  def download_path gf, options = {}
+    GenericFile.find(gf.id).doi_url || sufia.download_path(gf.id, options)
```
mbarnett: Few comments (I dug into this in my own branch because it bit me, too):

```ruby
def download_path gf, options = {}
```

This still won't work in all cases, because you're assuming download_path takes at most 2 arguments, the second of which must be a hash. In actuality:

```ruby
irb(main):001:0> Sufia::Engine.routes.url_helpers.method(:download_path).arity
=> -1
```
-1 indicates that the method has variable arity, so you need to match that definition:

```ruby
def download_path(gf, *more)
....
.... || sufia.download_path(gf.id, *more)
```

```ruby
GenericFile.find(gf.id).doi_url || ...
```
I see a couple of questions/issues here:

1) This is called above as

```ruby
download_path(@generic_file)
```
so gf is a GenericFile which means

```ruby
002:0> gf == GenericFile.find(gf.id) 
=> true
```
Is there any point to this find call? If there is, we should add a comment explaining the logic behind it, because at first glance it's redundant.

2) doi_url.present? will work better than just .doi_url if doi_url is ever an empty string

3) Is the assumption that everything passed into download_path will respond to doi_url safe? Sufia doesn't make this assumption in it's download_path code, so I'm a bit leary of it. Might be better to check with respond_to? before doing this.

pgwillia: Sufia::Engine.routes.url_helpers.method(:download_path).arity -- interesting.  I had trouble tracking down how sufia.download_path was declared.  Am I wrong that [ActionDispatch::Routing::PolymorphicRoutes#polymorphic_path](http://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html) is responsible?  

I'm new to arity so help me out.  Wouldn't ```def download_path(gf, *more)``` and  ```def download_path(gf, options = {})``` both have [arity of -2](http://ruby-doc.org/core-2.2.0/Method.html#method-i-arity)? I guess this implies that we have the case "For methods written in C, returns -1 if the call takes a variable number of arguments."

I'm happy to change it.  Both work for me.  The only place I see a second argument is calls of the form ```download_path(@generic_file, file: 'thumbnail')```

pgwillia: 1) I recall having casting problems without using GenericFile.find.  In some situations it was returning a SolrDocument (or something like that) which doesn't have the .doi_url attribute.

2) I want the method to return a url in either case.  

3) Maybe I'm assuming too much. My understanding is that my download_path method would return a doi_url if not falsey otherwise do the normal thing for sufia.download_path.

pgwillia: recall doi_url
```
      def doi_url
        self.identifier.first if ((!self.identifier.first.nil?) && (linkable? self.identifier.first))
      end
```

mbarnett: Re: Routes

It is not related to polymorphic_path. It's coming from Rails by way of Sufia's routes file. Specifically [once Sufia defines a :downloads resource](https://github.com/projecthydra/sufia/blob/master/config/routes.rb#L36), Rails kicks into gear and [generates named route methods for it](https://github.com/rails/rails/blob/51c7ac142d31095d4c699f44cc44ddea627da1eb/actionpack/lib/action_dispatch/routing/route_set.rb#L279-L289). All rails route helpers are just defined as taking *args, an indistinct number of arguments. Which explains why the arity is -1, not -2 (which is super ambiguous since this was not written in C).

Doing it this way is mostly futureproofing. download_path(a, b={}) and download_path(a, *b) behave identically, until sufia or anyone else changes the download path route definition to add an extra param. If that happens, you'll have to change your download_path override, whereas it should(knock-on-wood, no guarantees) keep working in the latter definition.

pgwillia: Cool.  I think that makes sense. Is it safe to assume that the first argument is a GenericFile if that's the case?

mbarnett: For 1 you might want to expand this logic and only do the find if it isn't a GenericFile but is a SolrDocument? I'm thinking mostly that ActiveFedora#find is sloooow and avoiding it when you can is good. 

For 2 I was thinking something alone the lines of:

```ruby
    gf.doi_url.present? ? gf.doi_url : sufia.download_path(gf.id, *more)
```

It looks like the linkable? call takes care of the empty string concern, though.

3) I guess my concern with the assumption is that we're shadowing the download resource route from Sufia here, and I guess my philosophy from a principle of least astonishment/liskov substitution principle point of view is that we shouldn't radically change the contract of that method; you can call Sufia's download_path with non-GenericFiles, but calling our shadowing function with a non-GenericFile will throw an exception

There's also a little tiny part of my that, however unlikely, can imagine someone calling download_path on some other type of object entirely that happens to respond to id with an id that coincidently matches a GenericFile's id, and then everything works entirely by accident in the worst possible way.

The format of ActiveFedora ids probably helps mitigate that, though

pgwillia:
I've updated the solution to be more cautious of types and the arguments structure.

mbarnett:
'tis a model of paranoia. Throw something at me for being nitpicky but you can remove a redundancy with

```ruby
gf && gf.doi_url.present? ? gf.doi_url : sufia.download_path(*args)
```

pgwillia:
DRYed.

mbarnett:
:+1:

Issue #595, opened 2015-09-04
Fix malformed XML in some batch-ingested FOXML in old ERA


pbinkley:
After Henry gives us a fresh dump of the complete ERA FOXML, we'll have to fix the items that Mariana has identified (see lists in https://drive.google.com/drive/folders/0Bx9c8G8_FeK5OTNwN3FDZERpYTg ). Rather than attempt to reinsert them into Fedora (with the risk that something will go wrong and cause downtime before the thesis deadline), we'll simply use the fixed FOXML in the migration. The migration script will have to be adjusted to reflect this (i.e. to avoid fetching the current FOXML from ERA, or to handle the error message that it will get if it tries).

pbinkley:
Migration script needs to catch up with this.

pbinkley:
Closed by #645

pgwillia:
Reopened by #646

PR #645, opened 2015-09-30
595 fix malformed xml

criedlberger:


pbinkley:
Closes #595 
( @criedlberger , you should add a line like this to the PR description, so that it will automatically close the issue when the PR is merged.)

PR #648, opened 2015-10-01
Migration: sort files, add completion log message

pbinkley:
Added a couple of tweaks to improve the logging of the big migration:

- sort the files, so they are processed in alphabetical order
- add file counter to log message ("file 104 of 903")
- add a completion entry that gives the uuid, file path and name, NOID, and communities and collections, so that we can grep out complete info

This should make it easier to monitor and if necessary restart the migration.

### DIFF
--- a/lib/tasks/migration.rake
+++ b/lib/tasks/migration.rake
@@ -216,10 +216,15 @@ namespace :migration do
     @ingest_batch = Batch.find_or_create(@ingest_batch_id)
     MigrationLogger.info "Ingest Batch ID #{@ingest_batch_id}"
     #for each metadata file in the migration directory
-    Dir.glob(metadata_dir+"/uuid_*.xml").sort.each do |file|
+    allfiles = Dir.glob(metadata_dir+"/uuid_*.xml")
+    filecount = allfiles.select { |file| File.file?(file) }.count
+    MigrationLogger.info "Files to process: " + filecount.to_s
+    thisfile = 1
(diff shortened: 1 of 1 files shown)


lib/tasks/migration.rake
```diff
@@ -216,10 +216,15 @@ namespace :migration do
     @ingest_batch = Batch.find_or_create(@ingest_batch_id)
     MigrationLogger.info "Ingest Batch ID #{@ingest_batch_id}"
     #for each metadata file in the migration directory
-    Dir.glob(metadata_dir+"/uuid_*.xml") do |file|
+    allfiles = Dir.glob(metadata_dir+"/uuid_*.xml")
+    filecount = allfiles.select { |file| File.file?(file) }.count
+    MigrationLogger.info "Files to process: " + filecount.to_s
+    thisfile = 1
+    allfiles.sort.each do |file|
```
mbarnett: not important, but a neat trick -- instead of defining a counter outside the loop, and manually incrementing it each iteration, you can do:

```ruby
allfiles.sort.each_with_index do |file, index|
...
end
```

The only caveat is that the indexing is zero-based.

pbinkley: Nice! I've just committed it.

mbarnett:
Running tests against this branch now

mbarnett:
It seems like there's no tests specifically broken on this branch that aren't broken on master. Merging.

PR #670, opened 2015-10-05
update hydranorth.rake to have a rake task that characterize files th…

weiweishi:
…at haven't been characterized.


pbinkley:
Should probably DRY it by setting an option on a single characterize task (e.g. true=recharacterize, false=only characterize if not already characterized). But I'm merging.

PR #671, opened 2015-10-05
fixes CCID for local Rails dev environment

mbarnett:
We broke this when setting up the real mappings for production with IST.

Like it says on the tin, this fixes it when Rails is run in development mode locally


pgwillia:
Implies that plano either a) isn't actually environment=development or b) won't have ualberta shib configured in the future.

mbarnett:
Since Plano and the Vagrants are configured differently, it might be an argument that they shouldn't be sharing a Rails environment?

mbarnett:
(this might also be an argument for running Plano in a different (non-development) Rails environment: https://github.com/ualbertalib/HydraNorth/issues/685. It won't correctly model some kinds of behaviour, like viewpath lookup, that we'll see in Production, because of Rails being annoying like that)

mbarnett:
I've got a better solution for this incoming on another branch

PR #672, opened 2015-10-05
fix issue with the server issuing 500s when records aren't found

mbarnett:
Right now the server 500s if you try to visit an item it can't find, eg) https://hydranorth.library.ualberta.ca/files/asdfg

This is suboptimal behaviour, so this commit rescues ActiveFedora and ActiveRecord not found errors properly so that we can render a human-friendly 404 page (and send the correct 404 status code to robots and other forms of life).

There's no issue associated with this, I just noticed it while fixing another issue, and it's annoying.


mbarnett:
WIP. Don't merge this yet. Waiting on wording from @sfbetz

sfbetz:
Language for 404 (we will need to update again after migration is complete):

The item you are looking for is not currently available in ERA. We are migrating our collections and it will take a few days for all the items to appear. Please contact us at erahelp@ualberta.ca if you would like more information about this migration.

mbarnett:
@sfbetz for your approval:

sfbetz:
Perfect! Thanks

sfbetz:
Wait! Footer is old footer still too.... will that update on production, or does some manual magic need to happen?

sfbetz:
Ignore last comment - forgot that we haven't yet deployed new content for footer...

mbarnett:
Ignoring like a pro

PR #682, opened 2015-10-06
Redirect url

pbinkley:
Restored Piyapong's branch to correct the issue with retrieving objects by their fedora3uuid.

### DIFF
--- a/app/controllers/redirect_controller.rb
+++ b/app/controllers/redirect_controller.rb
@@ -1,23 +1,23 @@
 class RedirectController < ApplicationController
   def item
     return render_404 ActiveRecord::RecordNotFound unless is_uuid
-    file = find_item
+    file = find_item_id
     return render_404 ActiveRecord::RecordNotFound unless file != nil
-    redirect_to "/files/#{file.id}", status: :moved_permanently
+    redirect_to "/files/#{file}", status: :moved_permanently
   end
 
   def datastream
     return render_404 ActiveRecord::RecordNotFound unless is_uuid && is_datastream
-    file = find_item
+    file = find_item_id
     return render_404 ActiveRecord::RecordNotFound unless file != nil
-    redirect_to "/downloads/#{file.id}", status: :moved_permanently
+    redirect_to "/downloads/#{file}", status: :moved_permanently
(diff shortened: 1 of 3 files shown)


app/controllers/redirect_controller.rb
```diff
@@ -39,21 +39,28 @@ def is_datastream
     return (/^DS\d+/ =~ params[:ds]) != nil
   end
 
-  def find_item
+  def find_item_id
     uuid = params[:uuid].split(":")[1]
-    file = GenericFile.find(fedora3uuid: uuid).first
```
weiweishi: It doesn't need to strip out the "uuid:" prefix for this solr query to work on the backend

app/controllers/redirect_controller.rb
```diff
@@ -39,21 +39,28 @@ def is_datastream
...
     render template: '/error/404', layout: "error", formats: [:html], status: :gone
   end
 
+  def find_id(uuid)
+    solr_rsp =  ActiveFedora::SolrService.instance.conn.get 'select', :params => {:q => Solrizer.solr_name('fedora3uuid')+':'+uuid}
+    numFound = solr_rsp['response']['numFound']
+    if numFound > 0
+      return solr_rsp['response']['docs'].first['id']
+    else
+      return nil
```
weiweishi: Still seeing problem with items that are not migrated yet.

app/controllers/redirect_controller.rb
```diff
@@ -1,23 +1,23 @@
 class RedirectController < ApplicationController
   def item
     return render_404 ActiveRecord::RecordNotFound unless is_uuid
-    file = find_item
+    file = find_item_id
     return render_404 ActiveRecord::RecordNotFound unless file != nil
```
weiweishi: unless file != nil should be changed to "if file.nil?" and that seem to fix the issue to redirect items that don't exist.

weiweishi:
@pbinkley @piyapongch , in order for us to make to the tagging deadline, I had to make some changes to the PR. All tests are passed and it works on production. But sorry this is not the proper way to merge a PR.

Issue #692, opened 2015-10-06
Collection size


pbinkley:
Our use style of collections in ERA is clearly overwhelming our Sufia-based model. I think we need to be able to render a collection with a Solr query to a collection facet rather than our current model of addressing all members. We made changes to support large collections in #201 but clearly not enough.

pgwillia:
Might be because the presenter queries fedora for each member of a collection to find permissions... which in the case of the "Northern North America Collection" is 1132 times.

    model.members.select do |item|
      render_context.can? :read, item
    end.count
[https://github.com/ualbertalib/HydraNorth/blob/master/app/presenters/hydranorth/collection_presenter.rb#L38-L40]

Quickest fix is to remove the count, or see if the value used in pagination is useful.

pbinkley:
Could we replace it with a solr query that searches for members of the
collection and returns a permissions facet with item counts? We could then
have logic to report a number to the user based on the user's permissions
to view different items.

pgwillia:
That's along the lines that I was thinking.  The query that is built to get the collection members already accounts for permissions. Using the numDocs from that or the value that used to calcuate pagination should accomplish this.  Just trying to figure out how to access either of those values from the presenter.

pgwillia:
I can see the value that I want in the render_context object (from [collection_presenter](https://github.com/ualbertalib/HydraNorth/blob/master/app/presenters/hydranorth/collection_presenter.rb)) but can't figure out the hash foo to isolate it.

```
...
"response"=>{
  ...
  "response"=>{
    "*numFound*"=>0,
    ...
  }
}
...
```

PR #703, opened 2015-10-08
removing total_items improves performance for large collections fixes…

pgwillia:
… #692

exclude long running test using: rspec --tag ~slow spec
only long running test using: rspec --tag slow spec

### DIFF
--- a/app/presenters/hydranorth/collection_presenter.rb
+++ b/app/presenters/hydranorth/collection_presenter.rb
@@ -3,9 +3,9 @@ class Hydranorth::CollectionPresenter < Hydranorth::Presenter
 
   self.model_class = ::Collection
   # Terms is the list of fields displayed by app/views/collections/_show_descriptions.html.erb
-  #temporarily remove size from the view to reduce loading time
+  #temporarily remove size and total_items from the view to reduce loading time
 
-  self.terms = [:title, :total_items, :description, :creator,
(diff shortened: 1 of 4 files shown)


pbinkley:
@pgwillia  I'm getting one failed test:
```
  1) CollectionsController index performance should be fast
     Failure/Error: expect( Benchmark.realtime{
       expected: < 0.5
            got:   3.30547144
```
Are you getting < 0.5 on your local vm, or can I ascribe this to the environment and merge?

pgwillia:
I was getting the expected performance on my VM (with 4 GB memory). 0.5 to 3.3 is a big difference there might be cause for concern there.

mbarnett:
Depending on what else the machine is running, what swap pressure is like, etc, I could see a 6x swing in performance pretty easily. A simple git stat on my vagrant can take anywhere from < 1 second to upwards of 30 depending on how warm everything is.

We may need to think up a better way of testing that performance, long term.

pbinkley:
Hmm, I'm using 2GB, which would probably explain it. I'll try again with 4GB.

pbinkley:
Oh, wait, no, it was 4GB.  I'm inclined to merge on the grounds that 3.3 is better than the error we were getting before. We can determine whether more work is needed when we see it on production hardware.

PR #720, opened 2015-10-13
take care of deprecation messages

pgwillia:


config/application.rb
```diff
@@ -28,6 +28,6 @@ class Application < Rails::Application
     # config.i18n.default_locale = :de
 
     # Do not swallow errors in after_commit/after_rollback callbacks.
-    #config.active_record.raise_in_transactional_callbacks = true
+    config.active_record.raise_in_transactional_callbacks = true
```
mbarnett: Have we actually tested that nothing breaks when we opt in to this new behaviour: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#error-handling-in-transaction-callbacks ?

I'm a bit reluctant to merge this and change the semantics of after_commit right before launch.

pgwillia: I have no problem with waiting on this.  Just wanted to get it off my VM before I forgot or wiped it clean.

mbarnett: Makes sense to me. Let's sit on this until after we're in prod.

pgwillia:
Minor annoyances.  Will create a new PR when config.active_record.raise_in_transactional_callbacks is thought through.

PR #723, opened 2015-10-13
697 migration permissions

criedlberger:


pbinkley:
@criedlberger could you look at the conflict when merging this with master? There's a use of is_part_of that I'm not sure about.

Issue #664, opened 2015-10-04
dynamically update stats number on the homepage


weiweishi:
Current the stats on the homepage is a hardcoded number:
![image](https://cloud.githubusercontent.com/assets/2047614/10268491/866b76b0-6a77-11e5-8c6e-f005bfe0b334.png)
Need to make it change dynamically with the actual content in the repo.

pbinkley:
The chart values are currently hard-coded at https://github.com/ualbertalib/HydraNorth/blob/master/app/assets/javascripts/custom.js#L78-L96 . We need a solr query that facets on item type for all accessible items (including ccid-protected but not private), in a json format suitable for this chart, which can be fetched by an xhr.

pbinkley:
Solr query: ```/select?q=active_fedora_model_ssi%3A%22GenericFile%22&rows=0&wt=json&indent=true&facet=true&facet.field=resource_type_sim```

PR #736, opened 2015-10-14
Add fresh hard-coded item type counts for front page

pbinkley:
As a first step to #664, these numbers reflect the current state of migration plus the legacy theses.


weiweishi:
the total count doesn't reflect the number with legacy thesis.

pbinkley:
No, I didn't update the total. I'll do that.

pbinkley:
And the other changes to custom.js shouldn't be there - must have messed up the copying between branches. Hang on.

pbinkley:
Squashed the messy commits. It's as mergable as it's going to get.

PR #749, opened 2015-10-15
Add fresh hard-coded item type counts for front page

pgwillia:
Redoing lost work.


pgwillia:
As a first step to #664, these numbers reflect the current state of migration plus the legacy theses.

PR #768, opened 2015-10-16
fixes #742 use dissertant for creator in the search result

### LINKED ISSUE #742: Theses need to display dissertant on search results page
Theses have no creator field; instead, the author is in the "dissertant" field. This is displayed correctly in the item-level page, e.g. here, where the creator "Smith, Lisa Wynne" comes from the dissertant field:

dissertant-item

But on the search results page the Creator is empty:

dissertant-search

We need to modify the search results to use creator field if it exists, otherwise use dissertant field if it exists.

weiweishi:


weiweishi:
updated the label so this PR is ready to merge.

PR #803, opened 2015-10-20
Add dynamic stats to homepage. Closes #664

### LINKED ISSUE #664: dynamically update stats number on the homepage

mbarnett:
Introduces a new JSON endpoint for repository statistics, refactors the repo stats javascript into its
own file (only included on the homepage), and includes some minor refactorings of the layouts and head partial


app/controllers/repository_statistics_controller.rb
```diff
@@ -0,0 +1,56 @@
+class RepositoryStatisticsController < ApplicationController
+
+  CHART_COMPONENTS = ["Report", "Image", "Thesis", "Journal Article (Published)"]
```
pbinkley: Sorry - the intention was not to pin these specific values, but to show the top four, whatever they may be. So the facet query should determine which four (+ "Other") get displayed.

mbarnett: Aha! I will refactor for that.

mbarnett:
Ok, I've changed the logic to call out the top 4 categories, and group the rest into "Other". This should be ready to merge now.

PR #821, opened 2015-10-27
Audit script

criedlberger:


lib/tasks/migration_audit.rake
```diff
@@ -0,0 +1,1030 @@
...
+  def check_collection(collection_attributes, model, communities, uuid)
+  
+     collection_id = find_object(uuid)
+     if collection_id == nil
+       AuditLogger.info "missing object: #{uuid}"
+       xml.uuid('uuid' => uuid) { 'Missing' }
+     else
+       @collection = Collection.find(collection_id)
+       current_user = User.find_by_username('admin')
+
```
pbinkley: We'll need the xml.uuid element here, like line 383

Issue #745, opened 2015-10-15
Use _ssim instead of _tesim fields to match collection ids


pbinkley:
See https://github.com/ualbertalib/HydraNorth/issues/729#issuecomment-148469217

Low priority because the probability of error is very low, but should get fixed.

pgwillia:
Fixed in #777

pgwillia:
Sorry, actually only half fixed: https://github.com/ualbertalib/HydraNorth/blob/master/app/models/search_builder.rb#L8

PR #822, opened 2015-10-27
fixes the solr query to include generic files that belongs to communi…

### LINKED ISSUE #814: Dataverse Community not showing its members
Items from Dataverse are now associated with a community called Dataverse Datasets (cf. #781), but the landing page for that community does not list any items.

weiweishi:
…ty directly, fixes #814

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -5,6 +5,6 @@ class SearchBuilder < Blacklight::SearchBuilder
 
    def include_collection_ids(solr_parameters)
     solr_parameters[:fq] ||= []
-    solr_parameters[:fq] << "(-hasCollectionId_tesim:[* TO *] AND belongsToCommunity_tesim:#{collection.id}) OR hasCollectionId_ssim:#{collection.id}"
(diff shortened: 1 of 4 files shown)


app/models/search_builder.rb
```diff
@@ -5,6 +5,6 @@ class SearchBuilder < Blacklight::SearchBuilder
 
    def include_collection_ids(solr_parameters)
     solr_parameters[:fq] ||= []
-    solr_parameters[:fq] << "(active_fedora_model_ssi:Collection AND belongsToCommunity_tesim:#{collection.id}) OR hasCollectionId_ssim:#{collection.id}"
+    solr_parameters[:fq] << "(-hasCollectionId_tesim:[* TO *] AND belongsToCommunity_tesim:#{collection.id}) OR hasCollectionId_ssim:#{collection.id}"
```
pbinkley: Could we fix the tesim fields and close #745 at the same time?

weiweishi:
updated the PR according to @pbinkley's suggestion.

pbinkley:
Note: adding ```:symbol``` is the first step to getting rid of some ```tesim``` fields that contain unparsable tokens. Once we do a full reindexing, we'll be able to remove references to the ```tesim``` versions of these fields.

PR #836, opened 2015-11-04
Google stats

### LINKED ISSUE #529: Import download stats from old ERA to Google Analytics for HydraNorth
Download data must be migrated from MySQL table into Google Analytics, so that it can drive the analytics display for migrated items.

- [Migration](https://support.google.com/analytics/answer/3191589?hl=en)

Are there other implications for GA of our re-using the era.library domain that we should consider? 

We could filter by bots (since the db contains the user agent) but this has communication implications (since download numbers will drop).

We know that import is easy; export from mysql should only be difficult if we do the bot filtering.

We will not attempt to migrate old ERA's Google Analytics data to match new ERA urls.

- [ ] test bot filtering
- [ ] determine the effect on download numbers of bot filtering, decide whether to implement
- [ ] determine whether information about the service change needs to be incorporated into the analytics display (e.g. "Data before Sept. 2015 was collected by a different method")
- [ ] import the data at migration time

criedlberger:
Closes #529


lib/tasks/migration.rake
```diff
@@ -854,6 +871,45 @@ namespace :migration do
...
+      index = file.rindex('/')
+      id = file[index+1..-1]
+      id.slice! ".txt"
+     
+      begin 
+        @generic_file = GenericFile.find(id)
+        if @generic_file == nil
+          MigrationLogger.info "Generic file not found: #{id}"
+        else
+          @generic_file.add_file(File.open(file), path: 'gastats', original_name: file, mime_type: 'texl/xml')
```
pbinkley: Could we add a check here, to prevent remigrating the stats if they're already there? We're still running migration jobs and we may well process the same items more than once.

app/models/concerns/hydranorth/generic_file/ga_stats.rb
```diff
@@ -0,0 +1,11 @@
+module Hydranorth
+  module GenericFile
+    module GaStats
```
pbinkley: A more descriptive name would make this easier to keep track of in the future (since these stats don't come from GA) - could we call it ERA1Stats, both in the ruby and the javascript.

app/views/generic_files/_flot_stats.js.erb
```diff
@@ -0,0 +1,155 @@
...
+              if(a[i] === a[j])
+                  a.splice(j--, 1);
+          }  
+      }  
+
+      return a;
+  }; 
+
+  var stats = <%= @stats.to_flot.to_json.html_safe %>;
+  var old_stats = <%= @generic_file.gastats.content.html_safe %>;
```
pbinkley: call this era1_stats for consistency

lib/tasks/migration.rake
```diff
@@ -854,6 +871,49 @@ namespace :migration do
...
+    MigrationLogger.info "Files to process: " + filecount.to_s
+    allfiles.sort.each_with_index do |file, thisfile|
+      start_time = Time.now
+      MigrationLogger.info "Processing the file #{file} (#{thisfile + 1} of #{filecount})"
+      index = file.rindex('/')
+      id = file[index+1..-1]
+      id.slice! ".txt"
+     
+      begin 
+        byebug
```
weiweishi: This was merged into master without removing. I'm going to remove it in a separate PR.

PR #882, opened 2015-12-01
remove fastercsv dependency for batch ingest

weiweishi:
remove the dependency for fastercsv to avoid the outdated shebang issue at deployment.


pgwillia:
What's the outdated shebang issue?

Also notice that #856 doesn't include documentation.

henryzhang87:
If in a file, especially a script file, it has a line
"#!/usr/local/bin/ruby" or anything,
RPM package builder will assume the target machine needs
/usr/local/bin/ruby in place, it not, the package will abort installation.

However, in our case, ruby is not installed at a location called
"/usr/local/bin/", it is somewhere else.


The temporary fix is to replace all the /usr/local/bin/ruby with the right
path during RPM build

lib/tasks/batch.rake
```diff
@@ -28,7 +27,8 @@ namespace :batch do
 
   def convert_csv_json(file)
     csv = CSV.open(file, :headers => true, :header_converters => :symbol, :converters => :all)
-    json = csv.to_a.map { |row| row.to_hash }
+    json = csv.map{ |x| x.to_h }
+    puts json
```
pgwillia: @weiweishi did you mean to leave this in?

pgwillia:
The other thing I'm still working out is if I should change the collection/community. Currently the test I prepared fails with
```
[ERROR 2015-12-17 13:40:43] FAILED batch ingest with
[ERROR 2015-12-17 13:40:43] ActiveFedora::ObjectNotFoundError
```

PR #883, opened 2015-12-03
add config so that fulltext indexing is disabled by default fixes #875

### LINKED ISSUE #875: Disable full-text indexing
Prevent uploaded files from being passed to Solr for content extraction and indexing.

Some items remain in "processing" state, which prevents the user from editing them. The intermediate cause is that the Batch retains status "processing". This is sometimes associated with errors in Solr's attempt to extract content.

See OTRS ticket 2015120210963 for background and examples.

pgwillia:


pgwillia:
I'm not sure if this got everything.  I'm seeing that the update/extract is still called on Solr when I upload files.

pgwillia:
Nm. Need to restart resque-pool.

PR #890, opened 2015-12-09
Closes #884

### LINKED ISSUE #884: Error when searching within a collection
When searching within a collection (e.g. using the search box on https://era.library.ualberta.ca/collections/44558r66n ), I get a 500 error. The log has:

```
Completed 500 Internal Server Error in 375ms (ActiveRecord: 0.3ms)

NoMethodError (undefined method `[]' for nil:NilClass):
  app/controllers/concerns/hydranorth/collections_controller_behavior.rb:11:in `show'
```
I'm guessing this is something introduced by #868 .

mbarnett:
Resolves errors when searching within collections. Collection Solr config is copied at run time within HydraCollections from the Catalog Controller, so default search fields must be defined at class instantiation time and not within a before request method


weiweishi:
verified failed tests are not related. Some of them are fixed in branch https://github.com/ualbertalib/HydraNorth/tree/fix-spec-tests. Will continue to work on tests as part of #686.

PR #892, opened 2015-12-14
876 official collection

criedlberger:


app/controllers/concerns/hydranorth/collections/selects_collections.rb
```diff
@@ -16,12 +16,27 @@ def find_communities_with_edit_access
...
   end
 
-  def find_communities(access_level = nil)
+ def find_collections(access_level = nil)
     # need to know the user if there is an access level applied otherwise we are just doing public collections
     authenticate_user! unless access_level.blank?
 
     # run the solr query to find the collections
-    query = collections_search_builder(access_level).with({q: 'is_community_bsi:true'}).query
+    query = collections_search_builder(access_level).with({q: '(-is_community_bsi:true AND -is_community_tesim:true) AND (is_official_bsi:true OR is_official_tesim:true)'}).query
```
pbinkley: Looking for boolean values in a tesim field can't be right. It all works with  just ```{q: '-is_community_bsi:true AND is_official_bsi:true'}```. Is there some other reason for the tesim fields?

app/controllers/concerns/hydranorth/collections/selects_collections.rb
```diff
@@ -16,12 +16,27 @@ def find_communities_with_edit_access
...
+    # return the user's collections (or public collections if no access_level is applied)
+    @user_collections = response.documents
+ end   
+
+ def find_communities(access_level = nil)
+    # need to know the user if there is an access level applied otherwise we are just doing public collections
+    authenticate_user! unless access_level.blank?
+
+    # run the solr query to find the collections
+    query = collections_search_builder(access_level).with({q: '(is_community_bsi:true OR is_community_tesim:true) AND (is_official_bsi:true OR is_official_tesim:true) '}).query
```
pbinkley: Same thing: tesim fields are unnecessary?

pbinkley:
This includes a necessary but unsatisfactory hack to detect false boolean values. This should be revisited at some point. (See #897 )

PR #894, opened 2015-12-14
add empty homepage#index view to override the heavy default sufia beh…

pgwillia:
…aviour

fixes intermittant ActiveFedora::ObjectNotFoundError


mbarnett:
I think you might also want to sublass Sufia::HomepageController and override #index to eliminate these calls? https://github.com/projecthydra/sufia/blob/6.x-stable/app/controllers/concerns/sufia/homepage_controller.rb#L16-L23

pgwillia:
I thought so too but it didn't seem to be necessary when I tried it without
changes to the controller.

PR #923, opened 2016-01-14
Correct description form param to pass array value

### LINKED ISSUE #922: description entered on the upload form doesn't stick
The description entered on the upload form doesn't save properly, while other metadata fields are saved. Editing form seems to be fine. 
screenshot - 16-01-14 - 11 05 52 am

mbarnett:
aka Paramater single-or-multi-valuedness is critically important to form validation and security permissions. Closes #922.

### DIFF
--- a/app/forms/hydranorth/forms/generic_file_edit_form.rb
+++ b/app/forms/hydranorth/forms/generic_file_edit_form.rb
@@ -5,12 +5,12 @@ module Hydranorth
       include HydraEditor::Form
       include HydraEditor::Form::Permissions
       self.required_fields = [:title, :resource_type, :language, :creator, :subject, :license, :belongsToCommunity ]
-      
+
 
       # This is required so that fields_for will draw a nested form.
       # See ActionView::Helpers#nested_attributes_association?
       #   https://github.com/rails/rails/blob/a04c0619617118433db6e01b67d5d082eaaa0189/actionview/lib/action_view/helpers/form_helper.rb#L1890
-      def permissions_attributes= attributes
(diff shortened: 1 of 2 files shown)


app/forms/hydranorth/forms/generic_file_edit_form.rb
```diff
@@ -5,12 +5,12 @@ class GenericFileEditForm < GenericFilePresenter
...
       include HydraEditor::Form::Permissions
       self.required_fields = [:title, :resource_type, :language, :creator, :subject, :license, :belongsToCommunity ]
-      
+
 
       # This is required so that fields_for will draw a nested form.
       # See ActionView::Helpers#nested_attributes_association?
       #   https://github.com/rails/rails/blob/a04c0619617118433db6e01b67d5d082eaaa0189/actionview/lib/action_view/helpers/form_helper.rb#L1890
-      def permissions_attributes= attributes
+      def permissions_attributes=(attributes)
```
weiweishi: Just curious why the parentheses are needed here?

mbarnett: They're not ;). I was debugging in that file and cleaned up the style while I was at it.

Generally speaking I think it's better if we're consistent about whether we use them or not, and most of the code base seems to use them.

weiweishi: Aha. Okay. I see. Yes I agree that we should keep the style consistent!

weiweishi:
Tested on Plano and it's working properly. No additional test is breaking. this closes #922

PR #933, opened 2016-01-19
571 ark minter

criedlberger:


app/jobs/batch_update_job.rb
```diff
@@ -77,6 +77,36 @@ def update_file(gf, user)
     end
     save_tries = 0
     begin
+      ark_identifier = Ezid::Identifier.create(id: Ezid::Client.config.default_shoulder + gf.id)
```
pbinkley: Shouldn't we use gf.ark_id here, instead of recreating the ARK from the shoulder and the NOID? We always want the update to go the ARK that was previously registered, and if we ever use different shoulders, this will break.

config/initializers/sufia.rb
```diff
@@ -64,11 +64,30 @@
...
+  }
+
+
+
   # please run rake db:seed to create the collections and restart httpd.
   # The collection IDs will be added here
   # In production it assumes that the collections will be available in the system at the time of deposit
 
 
+  config.cstr_collection_id = "gh93gz59r"
```
pbinkley: These collection ids shouldn't be committed

PR #934, opened 2016-01-19
929 ark update metadata

criedlberger:


spec/services/hydranorth/ezid_service_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper'
+
+describe Hydranorth::EzidService do
+  let(:user) { FactoryGirl.find_or_create(:jill) }
+  let!(:generic_file) do
+    GenericFile.new do |f|
+      f.id =  "888888888"
```
pbinkley: I'm still getting an "identifier already exists" error from this:

```
E, [2016-02-05T13:25:55.124559 #32638] ERROR -- : EZID CreateIdentifier -- error: bad request - identifier already exists
FI, [2016-02-05T13:25:55.496687 #32638]  INFO -- : EZID GetIdentifierMetadata -- success: ark:/99999/fk4888888888
```
Since this id will be used whenever the test is run, a hard-coded value won't work for f.id. How about generating a UUID?

PR #944, opened 2016-01-26
925 ark display

criedlberger:


app/views/generic_files/_show_actions.html.erb
```diff
@@ -0,0 +1,49 @@
...
+    </p>
+    <p>Export to:
+      <%= link_to 'EndNote', sufia.generic_file_path(@generic_file, format: 'endnote') %>
+      &nbsp;|&nbsp;
+      <%= link_to 'Zotero', sufia.static_path('zotero'), {id: 'zoteroLink', name: 'zotero', class: 'lightboxLink'} %>
+      &nbsp;|&nbsp;
+      <%= link_to 'Mendeley', sufia.static_path('mendeley'), {id: 'mendeleyLink', name: 'mendeley', class: 'lightboxLink'} %>
+    </p>
+    <div>Permanent url:</div> 
+    <div><a href=<%=request.base_url + '/id/' + @generic_file.ark_id %>><%= request.base_url + '/id/' %><br/><%= @generic_file.ark_id %></a></div>&nbsp
```
pbinkley: We need to be able to handle items that don't yet have ARKs (I get a "no implicit conversion of nil into String" error now). Since the rake task to add ARKs will take a long time to run, we'll have a gap between deploying the new release and having ARKs for everything, so we need to handle this case. We should omit the 'Permanent url:" and the ARK itself. Also: seems better to use paragraphs rather than divs, to match the other "action" blocks above.

Issue #435, opened 2015-06-18
Google Scholar indexing


awakaruk:
Story: 
As a researcher I want my colleagues to find the articles I deposited in ERA via a Google Scholar search. 

Note: best practices for maximizing repository indexing are noted in an OR 2015 presentation by Anurag Acharya here: http://www.or2015.net/wp-content/uploads/2015/06/or-2015-anurag-google-scholar.pdf

sfbetz:
This is a high priority issue for discovery and access to the content of ERA. Some discussion about Google Scholar on Hydra Community list:
https://groups.google.com/forum/#!searchin/hydra-community/google$20scholar/hydra-community/oOmTMIZ5bZ0/HQM1XAKGw8IJ

Right now some ERA content is being successfully indexed by GS  (about 5690 titles) https://scholar.google.ca/scholar?hl=en&q=site%3A.era.library.ualberta.ca&btnG=&as_sdt=1%2C5&as_sdtp=

I don't know what will happen to these citations now that we've migrated.

pbinkley:
If you sort by date instead of citation count you find that some of our
indexed items have no citations known to Google, so presumably they were
harvested from the site rather than just being picked up from citations. We
changed our url structure in old ERA to suit Google's wishes (see
https://docs.google.com/document/d/1t8LMGsYEQtkvWOSGHLXJDGJ-8fLVPgumrBqwWHQjkMY/edit#heading=h.xa6tnyj5zdye
, "Implications for ERA", point 1), and I think that helped. There is
probably more we could do:

- Sufia doesn't have sitemaps
- implement the year/month browse for recent items

But this work is hard to prioritize when there's no way to tell whether
it's having an effect on Google's harvesting.

sfarnel:
Definitely worth a larger conversation now that we have content moved over.

On Thu, Oct 22, 2015 at 1:49 PM, Peter Binkley <notifications@github.com>
wrote:

> If you sort by date instead of citation count you find that some of our
> indexed items have no citations known to Google, so presumably they were
> harvested from the site rather than just being picked up from citations. We
> changed our url structure in old ERA to suit Google's wishes (see
>
> https://docs.google.com/document/d/1t8LMGsYEQtkvWOSGHLXJDGJ-8fLVPgumrBqwWHQjkMY/edit#heading=h.xa6tnyj5zdye
> , "Implications for ERA", point 1), and I think that helped. There is
> probably more we could do:
>
[...]

Issue #850, opened 2015-11-13
Google Scholar Indexing (3): Browse by date


johnhuck:
Issue 3 of Epic #435.

With reference to UAL's [Google Scholar Strategy document] (https://docs.google.com/document/d/1t8LMGsYEQtkvWOSGHLXJDGJ-8fLVPgumrBqwWHQjkMY/edit#bookmark=id.jd4nolw32z5s).

Ensure that ERA content is crawl-able by ensuring that we have a suitable 'browse by date' interface, as recommended in the inclusion guidelines (and perhaps configure exclusions appropriately in the robot.txt file).

https://scholar.google.com/intl/en/scholar/inclusion.html#crawl

B. If your website has thousands of papers or more, the best way to make sure they're all discovered by the search robots is to provide a way to list them by the date of publication or the date of record entry. Other forms of browse interfaces, such as browse by author or by keyword, often generate more URLs than your website can deliver to the search robots in a reasonable amount of time.

C. For websites with more than a hundred thousand papers, we recommend that you create an additional browse interface that lists only the articles added in the last two weeks. This smaller set of webpages can be recrawled more frequently than your entire browse interface, which will facilitate timely coverage of your recent papers by the search robots.

weiweishi:
- [ ] requires a time-based browsable page, maybe link on the front page within the sitemap function. 
- [ ] a sitemap (depending on what do we want to include? it may require a cache for faster constructing the map)
- [ ] libhub version of the site? 
- [ ] schema.org (regular google crawling)

pgwillia:
Sufia has recently uploaded items featured on the front page.  
![screenshot from 2016-03-22 10-52-58](https://cloud.githubusercontent.com/assets/1220762/13960069/696307ec-f01c-11e5-8dda-56da1e6e6448.png)
Using the same logic it should be easy to re-introduce recently uploaded items to Hydranorth.  @leahvanderjagt  do we want this on the front page, somewhere else, or on it's own page?

leahvanderjagt:
If they are all equal in terms of how they would support Google Scholar
indexing then I think it would be good to have it on the front page
somewhere as long as it wouldn't slow down the front page loading. I wasn't
here for the decisions made for how the front page would look (and I really
like the way it looks now) so I'm going to talk to @sbetz about this before
providing further guidance.

On Tue, Mar 22, 2016 at 10:57 AM, pgwillia <notifications@github.com> wrote:

[...]

PR #950, opened 2016-02-02
pre-reqs for generating sitemap

pgwillia:
based on scholarsphere. Use `rake sitemap:generate` to create. Need to see how long it takes to generate sitemap at ERA scale and to schedule job to regenerate sitemap.xml. Related to #435


config/sitemap.rb
```diff
@@ -0,0 +1,14 @@
+Sitemap::Generator.instance.load(host: 'era.library.ualberta.ca') do
+  path :root, priority: 1, change_frequency: 'weekly'
+  path :catalog_index, priority: 1, change_frequency: 'weekly'
+  User.all.each do |user|
```
pbinkley: Since we're getting out of the user portfolio business, do we want to feature user home pages in the Google sitemap? @sfbetz

config/sitemap.rb
```diff
@@ -0,0 +1,14 @@
...
+    literal Sufia::Engine.routes.url_helpers.profile_path(user.to_param), priority: 0.8, change_frequency: 'daily'
+  end
+  read_group = Solrizer.solr_name('read_access_group', :symbol)
+  GenericFile.where(read_group => 'public').each do |f|
```
pbinkley: Pulling in the full AF object just to get the url is very expensive - could we make this a solr-only operation? I'm afraid in production this will be like the audit process and take days to run. @pgwillia

pgwillia: I'll take a look!

pbinkley:
While you're looking at a Solr-based solution for the sitemap, could you also think about using the same process to generate a date-based browse list of items, as described in https://scholar.google.com/intl/en/scholar/inclusion.html#crawl: "We recommend that the URL of every article is reachable from the homepage by following at most ten simple HTML links."
A structure that uses the creation date and generates a hierarchy based on year/month/day (and maybe hour, to break up big batch imports? Probably not worth it) would do the trick. It should include only the items that are included in the sitemap.

pgwillia:
Updated!

pbinkley:
After conversation with @sfbetz I've removed user pages from the sitemap, and I've also removed the /catalog route (since we block this in robots.txt).

pgwillia:
I know we're trying to protect ourselves from crawlers with the robots.txt, but could this also be a barrier to being indexed by Google Scholar?

pbinkley:
The Google analyst said they're cool with the date-based browse list (#850)
+ sitemap, so blocking /catalog won't affect Google Scholar indexing.

johnhuck:
There are directions regarding configuring the robots exclusion protocol to allow Google's robots to crawl. Maybe it's a secondary consideration for Google. See section 4.: https://scholar.google.com/intl/us/scholar/inclusion.html#crawl

PR #963, opened 2016-02-10
filter collections when community is selected fixes #887

### LINKED ISSUE #887: Collection field in item edit screen not updating to match community
Communities are now listed in the item edit screen (yay!), but when you select a community, the collections list doesn't update to match collections that are in that community, and instead lists lots of collections that are not in that community (not sure if it's listing all of them).

8b7bc811-f017-436c-9b08-7f02bc998d1d
.

pgwillia:
* copies batch behaviour

### DIFF
--- a/app/assets/javascripts/hydranorth_update_collections.js
+++ b/app/assets/javascripts/hydranorth_update_collections.js
@@ -1,5 +1,6 @@
-$(document).on("change", '.community-select', function(){ 
-          currentClass = $(this).attr('class').split(' ')
+$(document).on("ready change", function(){ 
+          if (!($('.community-select').attr('class'))) return;
+          currentClass = $('.community-select').attr('class').split(' ')
           var index
           $.each ( currentClass, function(i, v) {
             if (v.match(/community-group/)) { index = v}
@@ -9,7 +10,7 @@ $(document).on("change", '.community-select', function(){
             type: 'GET',
             dataType: 'script',
             data:  {
-              community_id: $("option:selected", this).val(),
+              community_id: $("option:selected", '.community-select').val(),
               index: index
             },
             complete: function() {}, 
--- a/app/controllers/concerns/hydranorth/batch_controller_behavior.rb
+++ b/app/controllers/concerns/hydranorth/batch_controller_behavior.rb
@@ -54,7 +54,7 @@ module Hydranorth
 
 
     def update_collections
-      @filtered_collections = Collection.where(belongsToCommunity: params[:community_id])
+      @filtered_collections = find_filtered_collections_sorted 
       @index = params[:index]
       respond_to do |format|
         format.js {}
--- a/app/controllers/concerns/hydranorth/collections/selects_collections.rb
+++ b/app/controllers/concerns/hydranorth/collections/selects_collections.rb
@@ -52,4 +52,20 @@ module Hydranorth::Collections::SelectsCollections
       d1.title <=> d2.title
     end
   end
+
+private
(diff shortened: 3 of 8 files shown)


pgwillia:
Tests were failing because I didn't consider the impact of the change to the script to load communities on ready for page where the community-select doesn't appear (all the javascripts are included with every page).  I've force pushed the fix so that the commit history is cleaner.

pgwillia:
Added collection drop down on batch and file edit is sorted by title based on ruby sorting solution https://github.com/ualbertalib/HydraNorth/blob/794c9c51852f676d71ac18ff1eb18562b7b87edf/app/controllers/concerns/hydranorth/collections/selects_collections.rb#L51-L52.  Also noted that sorting by title implemented in hydra-collections v7.0.0 [projecthydra/hydra-collections@e8e57e5], we're using v5.0.3

pgwillia:
@pbinkley  observed that with a larger number of collections/communities this fix is not very smooth.  I've been able to reproduce with a large # of communities.  The two things that cause me concern is a) any document ready state (any edit in any field, some change of focus, etc) results in the javascript call to update_collections and b) I didn't expect there to be an LDP call for each collection in the selected community -- this really slows things down and makes the ux confusing when the # of collections in a community is large (i.e. University of Alberta Libraries Licensed Resources) -- I thought this solution relied only on a Solr query.

pbinkley:
Maybe this needs to be looked at in conjunction with #969, then, since they both have to do with what appear to be unnecessary instantiations of Collection objects when dealing with the indexing of a GenericFile. I suppose the problem here is the need to check permissions, so that I get all the collections I'm entitled to deposit into, including ones that aren't public. Solr join?

pgwillia:
@pbinkley The permissions are stored in Solr and are part of the query for collections in a community. No join necessary. I don't *think* this is related to #969. The call I'm seeing in the logs is
```
Started GET "/batches/9p290936c/update_collections?community_id=&index=community-group-1&_=1455136872615"
 for 192.168.33.1 at 2016-02-10 13:42:40 -0700
Processing by BatchController#update_collections as JS
  Parameters: {"community_id"=>"", "index"=>"community-group-1", "_"=>"1455136872615", "id"=>"9p290936c"}
  Load LDP (94.2ms) http://127.0.0.1:8983/fedora/rest/dev/1r/66/j1/12/1r66j112r Service: 
47411920
  Load LDP (98.2ms) http://127.0.0.1:8983/fedora/rest/dev/28/01/pg/32/2801pg32c Service: 474119
20
  Load LDP (108.2ms) http://127.0.0.1:8983/fedora/rest/dev/f1/88/1k/88/f1881k888 Service:
 47411920
  Load LDP (86.2ms) http://127.0.0.1:8983/fedora/rest/dev/h9/89/r3/20/h989r3203 Service: 474119
20
  Load LDP (99.6ms) http://127.0.0.1:8983/fedora/rest/dev/7m/01/bk/68/7m01bk68h Service: 
47411920
  Load LDP (118.1ms) http://127.0.0.1:8983/fedora/rest/dev/dz/01/0q/04/dz010q04x Service: 47411
920
  Load LDP (103.5ms) http://127.0.0.1:8983/fedora/rest/dev/ws/85/9f/64/ws859f64s Service:
 47411920
  Rendered batch/_filtered_collections.html.erb (1932.6ms)
  Rendered batch/update_collections.js.erb (1933.1ms)
Completed 200 OK in 1943ms (Views: 1937.9ms | ActiveRecord: 0.8ms)
```

pbinkley:
Could it be that the problem is triggered by going through views/batch for
this? Could we simplify things by making this a view under
views/communities?

pgwillia:
This is ready

PR #976, opened 2016-02-19
974 fedora3uuid symbol

### LINKED ISSUE #974: fedora3uuid should not be tokenized
#947  reveals bug in the keyword search of uuids where extra items are included in the result set because fedora3uuid is being tokenized. A fix for this test is straightforward, but causes regressions in the dataverse migration and migration rake specs and requires reindexing for a proportion of the existing items.

pgwillia:
I think I may have been trying to change too many things before.  This closes #974


pbinkley:
Am I right in thinking that if we were to deploy this in production now, uuid searching would fail to find items that had not been reindexed? If so, is there any scaffolding we could add to make this work against both old-index and new-index items, and which we could remove after a full reindexing? If this is impracticle then I think we need to merge this (and potential other changes) into a branch which will only be merged into master and deployed when we are ready to perform that big reindexing (which has a queue of other updates which have to happen first: #977 ).

pgwillia:
You're thinking correctly. This PR would only find uuid's that exactly match item's fedora3uuid_ssim. We've established that while some items do have this field, most don't. Compare this to the current keyword search which would match the intended target, but potentially other spurious results as well. Since the PR results are a subset of the current master results I can't think of any useful scaffolding.

As long as we don't loose track of this, you can leave the PR hanging or merge to a branch. Does this PR need a label?

pbinkley:
Thanks. I think I'll merge to a branch, so we can call this issue done, and
add a note to #977 to keep track of the relationship of that branch to the
reindexing epic.

spec/tasks/migration_rake_spec.rb
```diff
@@ -69,7 +69,8 @@
       end
       Rake::Task.define_task(:environment)
       Rake::Task["migration:eraitem"].invoke('spec/fixtures/migration/test-metadata/multifile-metadata')
-      result = ActiveFedora::SolrService.instance.conn.get "select", params: {q:["fedora3uuid_tesim:uuid:846f544d-94db-41b4-9f4a-654e1457ed8c"]}
+      result = ActiveFedora::SolrService.instance.conn.get "select", params: {q:["#{Solrizer.solr_name('fedora3uuid')}:uuid:846f544d-94db-41b4-9f4a-654e1457ed8c"]}
+      byebug
```
pbinkley: whoops, gotta remove the byebug

pgwillia:
Removed byebug crumb.

pbinkley:
Merged to ```reindexathon``` branch, which won't be deployed until we're ready for the big reindexing.

PR #979, opened 2016-02-22
add logging in remove_lapsed_embargoes rake task

weiweishi:
* added setup_logger task to allow rake task to output logging information to the stdout. to use, switch => :environment with => :setup_logger so that the logging shows on the stdout for cronjob to capture
* add more logging information to the task.


lib/tasks/hydranorth.rake
```diff
@@ -6,15 +6,27 @@ namespace :hydranorth do
...
+  end
+
   desc "Remove lapsed embargoes"
-  task :remove_lapsed_embargoes => :environment do |t|
+  task :remove_lapsed_embargoes => :setup_logger do |t|
     items = Hydra::EmbargoService.assets_with_expired_embargoes
+    logger.info "Number of items that have expired embargoes: #{items.count}"
+    count = 0
     items.each do |item|
+      logger.info "clear expired embargo for #{item.id}"
```
pbinkley: how about a timestamp on this line? Then you could grep for the noid and get all the info on one line.

PR #1001, opened 2016-03-05
989 delete users

### LINKED ISSUE #989: User Management: Allow Admin to delete users
From #833. Provide this function on /users list. Confirm that there are no items for which this user is the depositor (i.e. it's up to admin to reassign items before deleting the user).

pbinkley:
Allow admin to delete other users. Closes #989.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -10,10 +10,6 @@ class UsersController < ApplicationController
   end
 
   def destroy
-# TODO 
-# - logout any sessions of this user
-
-    @user = User.from_url_component(params[:id])
     @email = @user.email
     if number_of_deposits(@user) == 0
       @user.destroy    
@@ -21,8 +17,7 @@ class UsersController < ApplicationController
     else
       flash[:error] = "User \"#{@email}\" cannot be deleted, since s/he has deposited items"
     end
-# TODO specify path properly
-    redirect_to '/users' 
+    redirect_to '/users/' 
   end
 
   def user_is_current_user
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -8,6 +8,13 @@ class UsersController < ApplicationController
     @trophies = @user.trophy_files
   end
 
+  def destroy
+    @user = User.from_url_component(params[:id])
+    @user.destroy    
+    flash[:success] = "User deleted"
+    redirect_to '/users' 
+  end
+
   def user_is_current_user
     redirect_to sufia.profile_path(@user.to_param), alert: "Permission denied: cannot access this page." unless @user == current_user || current_user.admin?
   end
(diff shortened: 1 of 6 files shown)


pbinkley:
WIP
To think about:
- users can't be deleted if there are any GenericFiles that have them as depositor. Should we check for other associations? (Collections don't matter, since now only admin can own them).
- can we add another level of "are you really really sure" confirmation before deleting
- can we use html tags in notices, to make the user id really really clear

spec/controllers/users_controller_spec.rb
```diff
@@ -68,4 +69,24 @@
...
     end
   end
+
+  describe "#destroy" do
+    let(:deletable) { FactoryGirl.create :alice }
+    before(:each) do
+      sign_in admin
+    end
+    describe "destroying users" do
+      it "should not delete user with GenericFiles" do
```
pbinkley: I need some help on this one: this spec fails (it deletes the user even though the user has a deposit, which should block it at https://github.com/ualbertalib/HydraNorth/pull/1001/files#diff-4e05ad0d64e6100656b63ad1e78f32c5R27 ). Why doesn't the controller detect the deposit?

app/controllers/users_controller.rb
```diff
@@ -21,6 +22,17 @@ def edit
     @trophies = @user.trophy_files
   end
 
+  def destroy
```
pgwillia: @pbinkley you had asked about how to handle user sessions.  I think this is what you need: ```sign_out :user``` https://github.com/plataformatec/devise/blob/master/lib/devise/controllers/sign_in_out.rb#L47-L67

spec/controllers/users_controller_spec.rb
```diff
@@ -68,4 +69,24 @@
...
+        @generic_file = GenericFile.create { |gf| 
+          gf.apply_depositor_metadata(user) 
+        }
+        expect( number_of_deposits(user) ).to eq(1)
+        expect { user.destroy }.not_to change(User, :count)
```
mbarnett: You're calling #destroy on the user *model*, but you created a method called ```destroy``` in the User _Controller_. You've got the right instinct in this test -- that is, that this is model behaviour that you want to enforce everywhere, not just in one controller -- but it's a model test, not a controller test, so it's in the wrong directory.

Fundamentally, I don't think you want that ```destroy``` code in the controller at all. What you want to do is create a ```before_destroy``` callback in the user *model* that returns false if-and-only-if the user has deposits. [The Rails Guide on the ActiveRecord Callback Lifecycle](http://guides.rubyonrails.org/active_record_callbacks.html) has a lot of good details (ActiveFedora copies the ActiveRecord Callbacks more-or-less verbatim).

pgwillia:
@pbinkley do you recall the status of this PR?

pbinkley:
I never got back to finish the tests - I tried to implement Matt's suggestions and couldn't make it work. I was in over my head at that point and wanted to find time to follow up on callbacks, but never did. We haven't had user deletion requests from the stakeholders, so I think this is low priority now - we've just been doing them through the console.

pgwillia:
Shall we close the PR then?

pbinkley:
Yes - closing because the work was abandoned.

PR #1002, opened 2016-03-07
Fix triple loading corruption

### LINKED ISSUE #969: Prevent triple-loading of collections when GenericFiles are reindexed
In the spike on Fedora API use #948 we found  that collections objects were being loaded three times during update_index - see jetty log:

```
INFO 14:52:47.788 (FedoraLdp) HEAD for: dev/j9/60/20/64/j9602064p
INFO 14:52:47.796 (FedoraLdp) GET resource 'dev/j9/60/20/64/j9602064p'
INFO 14:52:48.222 (FedoraLdp) HEAD for: dev/j9/60/20/64/j9602064p
INFO 14:52:48.231 (FedoraLdp) GET resource 'dev/j9/60/20/64/j9602064p'
INFO 14:52:48.684 (FedoraLdp) HEAD for: dev/j9/60/20/64/j9602064p
INFO 14:52:48.693 (FedoraLdp) GET resource 'dev/j9/60/20/64/j9602064p'
INFO 14:52:49.112 (FedoraLdp) HEAD for: dev/th/83/kz/38/th83kz38r/full_text
INFO 14:52:49.119 (FedoraLdp) HEAD for: dev/th/83/kz/38/th83kz38r/content
```

mbarnett:
Resolve issue where adding an object to a collection could result in the hasCollection attribute (normally the collection's title) being replaced with a hasCollection_ref that ActiveFedora does not cache and must load from Fedora (potentially multiple times) when loading an object.

Also includes a rake task to repair objects already impacted by this corruption, although the performance of the rake task against the production repository may not be satisfactory -- if anyone knows a better way to enumerate GenericFile Ids in the Fedora repo, we should use that instead of going through ActiveFedora.

Still investigating various means of testing the rake task.

Closes #969

### DIFF
--- a/lib/tasks/repair_collections_refs.rake
+++ b/lib/tasks/repair_collections_refs.rake
@@ -1,0 +1,19 @@
+namespace :hydranorth do
+  desc 'Repair objects whose hasCollection property has been replaced with hasCollection_ref'
+  
+  task :repair_collections_refs => :environment do
+    GenericFile.all.each do |file|
+      response, xml = Hydranorth::RawFedora.get(file.id, 'fcr:export', format: 'jcr/xml')
+      next unless response == 200
+      
+      unless xml.xpath('//sv:property[@sv:name="ns002:hasCollection_ref"]').empty?
+        c = Collection.find(file.hasCollectionId.first)
+        
+        # removing and re-adding the file should replace the file's collection_ref
+        # with the actual collection information
+        c.remove_member_id(file.id)
+        c.add_member_ids [file.id]
+      end
+    end
+  end
+end
--- a/app/models/concerns/hydranorth/collection_behavior.rb
+++ b/app/models/concerns/hydranorth/collection_behavior.rb
@@ -159,7 +159,7 @@ module Hydranorth
             end
           end
         else # I'm a collection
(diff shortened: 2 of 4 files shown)


lib/tasks/repair_collections_refs.rake
```diff
@@ -0,0 +1,19 @@
...
+  desc 'Repair objects whose hasCollection property has been replaced with hasCollection_ref'
+  
+  task :repair_collections_refs => :environment do
+    GenericFile.all.each do |file|
+      response, xml = Hydranorth::RawFedora.get(file.id, 'fcr:export', format: 'jcr/xml')
+      next unless response == 200
+      
+      unless xml.xpath('//sv:property[@sv:name="ns002:hasCollection_ref"]').empty?
+        c = Collection.find(file.hasCollectionId.first)
+
```
pbinkley: Could we add a logging statement here, so that we'll know which items were actually modified? I'm thinking we should deploy this to test and run the rake task and then check modified items (or at least spot-check) before we run this on production data.

mbarnett: That makes sense. I'll add some information to this and push an update.

PR #1013, opened 2016-03-10
Resolve collection select issue when changing communities

### LINKED ISSUE #1012: Edit form: selected collection vanishes after a second
In current master, in the edit form for an item: if you select a community, the collection dropdown updates to contain a filtered list (only the collections that belong to the selected community). But if you then select a collection, after a second or so the collection dropdown reverts to blank, as if you had not selected a collection. If you click update, the item is saved with no selected collection. I noticed this in a community with a single collection, but I've tested it in a community with two collections, and the bug still occurs. 
This doesn't happen in the current tagged release  v1.2.9 (in production), so I'd like to fix this before we deploy the new release.

mbarnett:
Fixes an issue where the ajax call to refresh the collection options was running on every page change, instead of only when the community select changes. Further cleanup of this file may be warranted (the index code doesn't seem to serve a purpose, and if it does, should likely leverage a custom data attribute instead), but this doesn't impact the immediate bug. Closes #1012.

### DIFF
--- a/app/assets/javascripts/hydranorth_update_collections.js
+++ b/app/assets/javascripts/hydranorth_update_collections.js
@@ -32,7 +32,7 @@ function update_collections() {
 
 
 $(function() {
-	// ensure collection optionss match selected community on page load
+	// ensure collection options match selected community on page load
 	update_collections();
 	
 	$('#generic_file_belongsToCommunity').change(function() {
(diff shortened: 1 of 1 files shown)


pbinkley:
OK I'll test and merge now

PR #1019, opened 2016-03-14
969 triple loading corruption

mbarnett:
Re-oppening against sprint branch, but adding logging to the rake task is still on my TODO list, so hold off on merging.

### DIFF
--- a/lib/hydranorth/raw_fedora.rb
+++ b/lib/hydranorth/raw_fedora.rb
@@ -1,0 +1,39 @@
+module Hydranorth
+  module RawFedora
+    
+    def get(id, subresource = '', additional_arguments = {})
+      response = perform :get, to_pair_tree(id) + subresource, additional_arguments
+    
+      xml = Nokogiri::XML.parse(response)
+      return response.code, xml
+    end
+    
+    def perform(action, path, arg_hash = {}, request_headers = {})
+      request_headers.reverse_merge!({cache_control: :no_cache})
+      user = ActiveFedora.config.credentials[:user]
+      password = ActiveFedora.config.credentials[:password]
+      endpoint = ActiveFedora.config.credentials[:url] + ActiveFedora.config.credentials[:base_path]
+      
+      url = endpoint + path + stringify_args(arg_hash)
+      
+      RestClient::Request.execute(
+        method: action, 
+        url: url,
+        user: user,
+        password: password,
+        headers: request_headers)
+    end
+    
+    def to_pair_tree(id)
+      "/#{ActiveFedora::Noid.treeify(id).strip}/"
+    end
+    
+    def stringify_args(arg_hash)
+      args = arg_hash.empty? ? '' : '?'
+      arg_hash.each { |k, v| args << k.to_s + '=' + v.to_s }
+      args
+    end
+    
+    module_function :get, :perform, :to_pair_tree, :stringify_args
+  end
+end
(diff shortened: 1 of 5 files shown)


pbinkley:
I'm getting this, even after ```bundle install``` - what am I missing?

```
/usr/lib64/ruby/gems/2.1.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:274:in `require': cannot load such file -- rest-client (LoadError)
	from /usr/lib64/ruby/gems/2.1.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:274:in `block in require'
	from /usr/lib64/ruby/gems/2.1.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:240:in `load_dependency'
	from /usr/lib64/ruby/gems/2.1.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:274:in `require'
...
```

mbarnett:
Looks like a missed dependency in the Gemfile

mbarnett:
Should work now

lib/hydranorth/raw_fedora.rb
```diff
@@ -0,0 +1,39 @@
...
+      "/#{ActiveFedora::Noid.treeify(id).strip}/"
+    end
+    
+    def stringify_args(arg_hash)
+      args = arg_hash.empty? ? '' : '?'
+      arg_hash.each { |k, v| args << k.to_s + '=' + v.to_s }
+      args
+    end
+    
+    module_function :get, :perform, :to_pair_tree, :stringify_args
```
pgwillia: I'm not familiar with module_function.  What does it do?

mbarnett: When you declare a method in a Module, it's an instance method. So,


```ruby
class Foo
  include Hydranorth::RawFedora
end
```
makes the methods available to *instances* of Foo, but you can't call them directly on the class or module. ```module_function``` declares Module-level aliases for the methods, so that you can call ```Hydranorth::RawFedora#get``` etc directly on the module.

You could declare them as module-level functions directly, but then they wouldn't become instance methods when mixed in to classes. It's a best-of-both-worlds thing. http://apidock.com/ruby/Module/module_function

pgwillia:
Seeing failing spec:
```
Failures:

  1) Collection should not insert a hasCollection_ref instead of hasCollection
     Failure/Error: expect(xml.xpath('//sv:property[@sv:name="ns002:hasCollection"]')).not_to be_empty
       expected `[].empty?` to return false, got true
     # ./spec/models/collection_spec.rb:98:in `block (2 levels) in <top (required)>'

Finished in 22.11 seconds (files took 21.72 seconds to load)
13 examples, 1 failure

Failed examples:

rspec ./spec/models/collection_spec.rb:93 # Collection should not insert a hasCollection_ref instead of hasCollection
```

I'll see if I can figure out why.

pgwillia:
Looks like its a problem with a different namespace?
```
<sv:property sv:name="ns005:hasCollection" sv:type="String" sv:multiple="true">
    <sv:value>A title</sv:value>
  </sv:property>
```
vs 
```expect(xml.xpath('//sv:property[@sv:name="ns002:hasCollection"]')).not_to be_empty```

mbarnett:
Works on My Machine™

(looks like the namespaces are machine-dependent).

mbarnett:
Pushed a fix that should find the right namespace prefix on any machine. Give it another shot when you have a second.

PR #1020, opened 2016-03-14
Only render collections that actually belong to the selected community

### LINKED ISSUE #1016: Collections disappearing from edit page
I think that an old bug has resurfaced. Collection appears to be dropping out of the edit page. See here: https://era.library.ualberta.ca/files/bg257h774/edit

This item is in the Daily Record collection, but on edit, collection disappears from that field. :(

Sounds like #916 has reappeared?

mbarnett:
Closes #1016. Eliminates the need for an AJAX call immediately on page load. Also refactors SelectsCollections's logic for finding collections based on a community ID, so that the code is more composable and does not rely on the presence of a particular request parameter.


mbarnett:
wrong branch

Issue #981, opened 2016-02-23
Ingest updated Dataverse metadata (Feb 2016)


johnhuck:
We are using the occasion of preparing an ingest of updated Dataverse metadata into ERA as an opportunity to address several outstanding Dataverse issues.

If possible, we would like to coordinate the timing of the metadata harvesting with @wyman10 so that it captures several Dataverse studies that are close to being released.

Once these outstanding issues are resolved, we can proceed with the harvesting sequence.
- [x] #450 Dataverse object Deposit/modified time
- [x] #674 Author Order between Dataverse and HydraNorth
- [x] #951 Dataverse update should not create duplicates
- [x] #547 Add 'Abstract' to metadata display [resolved]

Harvesting sequence:
- [x] @wyman10 emails @johnhuck with confirmation that harvesting can begin.
- [x] @johnhuck creates ticket in Dataverse repo to request export
- [x] @piyapongch asks Henry to clear all files from the export destination folder(s) and then runs export job.
- [x] as required by outcome of #450, @johnhuck creates a ddi->dcterms stylesheet 
- [x] @johnhuck transforms exported metadata
- [x] @weiweishi runs script to ingest or update Dataverse objects in ERA.

Other Dataverse issues (open issues, but no dependencies for harvesting):
- #704 HTML Markup being displayed in dataverse metadata and collection pages
- #493 Dataverse to HydraNorth pt. 2: automation

johnhuck:
After speaking with @sfarnel and @pbinkley, I have the understanding that the consensus is that we will attempt to capture the version dates from Dataverse for use as date uploaded and date modified in ERA, and so will prepare a new stylesheet to make dcterms metadata from the DDI.

johnhuck:
@sfarnel @sfbetz @wyman10 @weiweishi 

I have finished work on the DDI to dcterms stylesheet, which generates the same metadata for ERA as the old stylesheet, just from a different source. I have made a couple of modifications, though, that I wanted to run by you first. Let me know if you have any objections.

The first mod is the addition of dcterms:dateSubmitted and dcterms:modified, which precipitated the work on a new stylesheet in the first place.

[...]

Issue #951, opened 2016-02-03
Dataverse update should not create duplicates


pbinkley:
When importing from Dataverse, use DOI to detect previously imported items and update them rather than creating new records (which would create duplicates). This can be modeled on the detection of duplicate UUIDs in the migration.rake.

pbinkley:
Closed by #1043

PR #1043, opened 2016-04-02
674 951 dataverse ingest

### LINKED ISSUE #674: Author Order between Dataverse and HydraNorth
This appears to be an issue again in the latest import after being apparently resolved in a previous issue.

https://github.com/ualbertalib/HydraNorth/issues/555

Current example:
https://hydranorth.library.ualberta.ca/files/f4752g834#.VhL_NSpViko

Incidentally, if the same remedy can be applied to Subject order that might not be a bad thing.

weiweishi:
fixes #674, #951, and ready for #981


pbinkley:
I'm getting one failed spec:

```
  1) Migration rake tasks migration:dataverse_objects should update the metadata and not created new record
     Failure/Error: id = doc["id"]
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./spec/tasks/dataverse_migration_rake_spec.rb:79:in `block (3 levels) in <top (required)>'
```

Looking into it now.

pbinkley:
Further: I tried ingesting the test Dataverse object with: 
```
bundle exec rake migration:dataverse_objects['spec/fixtures/migration/test-metadata/dataverse']
```
Then I modified that record (I added another author) and ingested it again with the same command. It produced a new object with a new noid.

I'm looking into why ```duplicated?``` failed to find the first object. If I replicate the Solr query in the web interface, it finds both now.

weiweishi:
@pbinkley pushed a fix to the bug you identified.

pbinkley:
The new problem was [here](https://github.com/ualbertalib/HydraNorth/pull/1043/files#diff-b8835142c0b1c24a9a4fe38907c719faR238): my failed attempts had created 3 records with the same DOI, so the ```== 1``` was failing. I suppose we should log an error if there is more than one. I've deleted the extras and now I'm getting a different error, which I'll look into tomorrow.

weiweishi:
@pbinkley pushed a new PR to catch the error message you identified above, as well as the issue of adding to community member_ids (we no longer use member_ids as the information is captured in object's solr record.)

pbinkley:
I've added an explicit test to ensure that the duplicate didn't create a new record. I had to add [this](https://github.com/ualbertalib/HydraNorth/pull/1043/files#diff-d9a8b33daf55e89fa64cbdbd3de0290fR19) at the beginning of the spec to make sure that broken records from previous runs don't look like duplicates of the imported record. Since @wshi is away, and this is a small change limited to the spec, I'll go ahead and merge as is.

PR #1047, opened 2016-04-07
Remove GenericFileBatch#processing? calls

### LINKED ISSUE #1039: Sufia::GenericFile::Batches#processing? is sloooow
The main (only?) use-case for this at the moment is to determine whether or not to show the spinner beside an item when it has not yet been characterized after upload. Since this is only shown on a handful of items at any given time, this is a high price to pay when rendering every item in a list. Stakeholders have indicated a willingness to simply not display the spinner in any case for the time being.

Longer term solutions to this might include adding a special visibility or other property to the item's solr record on upload, and removing it once the item is characterized. Checking for this property would be much cheaper than this ```processing?``` call

mbarnett:
Performance enhancement. Closes #1039.


app/controllers/application_controller.rb
```diff
@@ -12,6 +12,8 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  around_filter :profile, if: -> {Rails.env.development? && params[:trace] == "1"}
```
weiweishi: What does around_filter do?

app/controllers/application_controller.rb
```diff
@@ -22,9 +24,24 @@ class ApplicationController < ActionController::Base
      render_404 exception
   end
 
+  def profile
```
weiweishi: How do we use this profiling tool on the app? Mind putting together a little doc on developers-handbook?

pbinkley: :+1:

app/controllers/application_controller.rb
mbarnett: Before filters run code before running the current route's action. After filters fun code after the current route's action. Around filters are a bit trickier in that they run it *around* the action, meaning both before and after.

The basic idea here is that before it starts running the action, it runs ```profile``` up to the point where the profile method calls ```yield```, at which point it runs the action. Once the action is done, it returns to profile and keeps running from the point ```yield``` was called.

So, in this case, it runs

```ruby
require 'ruby-prof'

RubyProf.start

yield
```

This starts the profile recording. Then it runs the current action for the route. And then after the action is finished, it runs the rest of the method, which stops the profiler's recording, and prints out a graph of the results

```ruby
results = RubyProf.stop

File.open "#{Rails.root}/profile-stack.html", 'w' do |file|
  RubyProf::CallStackPrinter.new(results).print(file)
end
```

There's more explanation in [the Rails Guide for ActionController](http://guides.rubyonrails.org/action_controller_overview.html#after-filters-and-around-filters).

app/controllers/application_controller.rb
mbarnett: On a dev machine, anywhere in the app, add a query param of ```trace=1```.

So for example if you want to generate a profile for http://localhost:3000/dashboard/files, you would visit http://localhost:3000/dashboard/files?trace=1. The page will (very slowly) load, and once it's done there will be a file called ```profile-stack.html``` in hydranorth folder with a graph of the profiler's results.

I'll update the handbook with this information too.

app/controllers/application_controller.rb
weiweishi: Thanks! Learnt something new today! :+1:

app/controllers/application_controller.rb
weiweishi: This is extremely useful. Thank you! @mbarnett

PR #1060, opened 2016-04-18
986 ualterms uri

pbinkley:
Changes needed for to fix the ualterms uris to match current data dictionary

### DIFF
--- a/bin/fix/fix.rb
+++ b/bin/fix/fix.rb
@@ -1,0 +1,143 @@
+require 'rest-client'
+# we use active_fedora/noid to generate pairtrees from noids
+require 'active_fedora/noid'
+
+require 'rdf/turtle'
+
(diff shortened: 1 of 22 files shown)


bin/fix/fix.rb
```diff
@@ -0,0 +1,144 @@
...
+					else
+						puts "Needs fixing: " + statement.object.to_s
+						fixable += 1
+						handleNoid(id, instance)
+					end
+				end
+			end
+		end
+	rescue => e
+  		e.response
```
mbarnett: Did you mean to print this out?

mbarnett:
I think this looks workable other than the possible missed puts noted above, and the issue I emailed you about.

pbinkley:
I've fixed the issues and force-pushed a single rebased commit.

Issue #792, opened 2015-10-19
Mouse-over on item title


pbinkley:
![odd-mouseover](https://cloud.githubusercontent.com/assets/272816/10567843/a7846f96-75cb-11e5-8ba0-85c98349c555.png)

Mousing over an item title on search results leads to partial removal of underlining. Should be complete.

johnhuck:
I've noticed that no matter how you resize the width of the browser window the remaining underline always adds up to a one complete line when it's split over two lines. Maybe that's a clue? Also it would be great if the descenders weren't cut off (e.g., 'g' 'p' '3').

pbinkley:
It seems that when the title is split over more than one line, in a mouseover the second line expands its top border or margin to cover the bottom of the first line, including the descenders and the underline. Any part of the first line that extends beyond the end of the second line is unaffected.

pbinkley:
Fixed by #1068

PR #1068, opened 2016-04-22
fixes an issue with lines being too squished when hovering over links

mbarnett:


mbarnett:
@pbinkley as discussed in sprint wrap-up meeting. Easy as pie.

pbinkley:
Closes #792

PR #1076, opened 2016-04-26
850 browse by date

pgwillia:
* replace twitter tile on front page with most recent public upload
* recent target shows last two weeks of content by default takes year and year/month parameters.
* older documents can be reached by links to month date buckets
* fix terms.library.library.ca namespace leftover
* if you have a bunch of failing spec on sign_in make sure you're using spec_helper NOT rails_helper!

### DIFF
--- a/app/controllers/recent_controller.rb
+++ b/app/controllers/recent_controller.rb
@@ -14,12 +14,14 @@ class RecentController < ApplicationController
     def recent
       params.permit(:year, :month)
       # grab any recent documents
-      (_, @recent_documents) = search_results({q: "#{Solrizer.solr_name('system_create', :stored_sortable, type: :date)}:#{date_range}", sort: sort_field, rows: 100}, search_params_logic)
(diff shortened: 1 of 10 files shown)


pbinkley:
Could we add a [```noindex```](https://developers.google.com/webmasters/control-crawl-index/docs/robots_meta_tag?csw=1#using-the-robots-meta-tag) header to these pages:

```
<meta name="robots" content="noindex" />
```

This will tell search engines that they can crawl these pages for links (because ```nofollow``` is not set) but not show these pages as search results.

pbinkley:
The buckets by month will continue to accumulate in the page, right? Ours will go back to 2009 when the first DSpace items were added. Could we improve the layout by grouping them by year?

```
<div>2016: January February ...</div>
<div>2015: January February ...</div>
...
```

And so on. (With only the months that actually have items appearing.)

pgwillia:
I can remove the zeros easy.  I actually intended on grouping them by year but punted when it was preventing me from getting this out.

PR #1083, opened 2016-05-06
Audit fix

weiweishi:
This PR includes:
* audit-fix script that address #977. Documentation is available on README. 
* rake tasks to handle index jobs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,3 +125,20 @@ Configuring Shibboleth
 * visit '/Shibboleth.sso/Metadata' to download and review the metadata, rename to a filename of your preference
 * upload the SP metadata to http://testshib.org/register.html or your Identiy Provider (IdP)
 * login by clicking on 'Sign into Shibboleth' and choosing one of the available identities
+
+Audit Fix
+---
+* To fix data with triple files:
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'triples','sample-sparql.txt'```
+* To remove irrelevant theses dates:
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'thesis'```
+* To fix reindex:
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'reindex'```
+A set of rake tasks is also added for index jobs:
+rake hydranorth:solr:index[id]         Index a single object with ID
+rake hydranorth:solr:index_pairtree[input]   Index with a pairtree structure
+rake hydranorth:solr:update_generic_file_index  Index all Generic File objects
+
+/bin/fix/fix.rb is to update all the namespace uris
+/bin/fix/run.sh is to run script through all the pairtree combinations. 
+
--- a/README.md
+++ b/README.md
@@ -131,14 +131,16 @@ Audit Fix
 * To fix data with triple files:
 ```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'triples','sample-sparql.txt'```
 * To remove irrelevant theses dates:
-```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'thesis'```
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'thesis' <pair tree prefix|id>```
 * To fix reindex:
-```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'reindex'```
-A set of rake tasks is also added for index jobs:
-rake hydranorth:solr:index[id]         Index a single object with ID
-rake hydranorth:solr:index_pairtree[input]   Index with a pairtree structure
-rake hydranorth:solr:update_generic_file_index  Index all Generic File objects
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'reindex' <pair tree prefix|id>```
 
-/bin/fix/fix.rb is to update all the namespace uris
-/bin/fix/run.sh is to run script through all the pairtree combinations. 
+A set of rake tasks is also added for index jobs:
+* ```rake hydranorth:solr:index[id]```         Index a single object with ID
+* ```rake hydranorth:solr:index_pairtree[input]```   Index with a pairtree structure
+* ```rake hydranorth:solr:update_generic_file_index```  Index all Generic File objects
+
+A shell script will update namespace uris
+* ```/bin/fix/fix.rb``` is to update all the namespace uris. Requires user to replace @server with the Fedora server location before using.
+* ```/bin/fix/run.sh``` is to run script through all the pairtree combinations. Requires being run from the bin/fix directory.
 
--- a/bin/fix/fix-audit.rb
+++ b/bin/fix/fix-audit.rb
@@ -89,10 +89,10 @@ def find_objects(input)
           id = statement.object.relativize(statement.object.parent).to_s
  	  if id.include? "-"
             puts "Not a noid: " + id
-	    notfixable += 1
+	    @notfixable += 1
           else
 	    puts "An ERA Object: " + statement.object.to_s
-            fixable += 1
+            @fixable += 1
             objects << id
           end
         end
@@ -144,7 +144,7 @@ end
 
 def reindex_single(id)
   start = Time.now
-  ActiveFedora::Base.find(args[:id]).update_index
+  ActiveFedora::Base.find(id).update_index
   finish = Time.now
   used_time = finish - start
   puts "reindexed #{id} used #{used_time}"
--- a/bin/fix/fix-audit.rb
+++ b/bin/fix/fix-audit.rb
@@ -18,7 +18,7 @@ def patch(path, payload)
   puts 'Patching ' + path + ': ' + payload
   response = RestClient::Request.execute(
     method: :patch,
-    url: @rest+path,
+    url: path,
     user: @user,
     password: @password,
     payload: payload,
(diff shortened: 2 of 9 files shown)


pgwillia:
@weiweishi remind me what we were going to do with this PR?  Is it going directly against master? Is it ready?

weiweishi:
I believe this is ready but I remember @pbinkley had a reason to not merging it at that time. @pbinkley could you please comment?

pbinkley:
I don't recall what I thought the issue was - maybe I was concerned about namespaces while we were tracking that down. Anyway, I see no reason not to merge.

bin/fix/fix-audit.rb
```diff
@@ -0,0 +1,188 @@
...
+  rails_env = Rails.env
+  config = YAML.load_file("config/fedora.yml")
+  @user = config[rails_env]['user']
+  @password = config[rails_env]['password']
+  @rest = config[rails_env]['url']
+  @base_path = config[rails_env]['base_path']
+end
+
+def patch(path, payload)
+  puts 'Patching ' + path + ': ' + payload
```
pgwillia: Are the puts statements meant to act as logging, in which case can we use a logger?

mbarnett: @pgwillia since this is a raw script with no active Rails environment or Rakefile logger available, I'm curious what the value of that would be versus just tee'ing stdout into a file when this script is run?

pgwillia: @mbarnett I guess I didn't think about it too deeply. Lots of puts usually means that using a logger is the next step. Realistically your solution would have the same result - with more chance for the operator to forget to pipe to file. I'm assuming that we will want to save this output for its potential to help us debug future prolems. But maybe we don't need to be worried about this here.

mbarnett: @pgwillia Mostly I'm thinking without a logger coming for free from Rails or Rake, it's YAGNI work just to set one up unless you know we GNI™

weiweishi: When the script is running, it does create a file for logging:
@f = File.open('audit-fix.log', 'a+'), and the information worth keeping has been written to this log file with @f.puts. -- @f.puts "Patched #{noid} with #{payload}: #{response}"

README.md
```diff
@@ -125,3 +125,20 @@ Configuring Shibboleth
 * visit '/Shibboleth.sso/Metadata' to download and review the metadata, rename to a filename of your preference
 * upload the SP metadata to http://testshib.org/register.html or your Identiy Provider (IdP)
 * login by clicking on 'Sign into Shibboleth' and choosing one of the available identities
+
+Audit Fix
+---
+* To fix data with triple files:
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'triples','sample-sparql.txt'```
+* To remove irrelevant theses dates:
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'thesis'```
```
pgwillia: Suggestions on how to test these? Obviously my problem is that I don't have the right data loaded.
```
bin/fix/fix-audit.rb:73:in `find_objects': undefined method `length' for nil:NilClass (NoMethodError)
	from bin/fix/fix-audit.rb:111:in `fix_theses_related_dates'
	from bin/fix/fix-audit.rb:180:in `<main>'
```

README.md
```diff
@@ -125,3 +125,20 @@ Configuring Shibboleth
...
+* To fix reindex:
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'reindex'```
+A set of rake tasks is also added for index jobs:
+rake hydranorth:solr:index[id]         Index a single object with ID
+rake hydranorth:solr:index_pairtree[input]   Index with a pairtree structure
+rake hydranorth:solr:update_generic_file_index  Index all Generic File objects
+
+/bin/fix/fix.rb is to update all the namespace uris
```
pgwillia: Where is @server defined?

pgwillia: Seems to expect bin/fix/fix.rb to be in the current directory. Am I invoking this wrong?

```
[root@hydranorthdev hydranorth]# ./bin/fix/run.sh 
ruby: No such file or directory -- fix.rb (LoadError)
...
```

README.md
```diff
@@ -125,3 +125,20 @@ Configuring Shibboleth
...
+/bin/fix/run.sh is to run script through all the pairtree combinations.
```
pgwillia: Seems to expect bin/fix/fix.rb to be in the current directory. Am I invoking this wrong?

```
[root@hydranorthdev hydranorth]# ./bin/fix/run.sh 
ruby: No such file or directory -- fix.rb (LoadError)
...
```

README.md
weiweishi: Do you have a thesis object that were ingested through migration script? You could use the spec fixture file for thesis and use the migration.rake to ingest it.

README.md
weiweishi: Where is it used? @pbinkley has hardcoded some of these values in bin/fix/fix.rb, and I might have missed it. The value should now all be read from the environment yml file, but I didn't go back and check his script.

pgwillia: @weiweishi it's in bin/fix/fix.rb

README.md
weiweishi: This bash script was when @pbinkley run the fix.rb on his local workstation. I think he probably ran it in the ./bin/fix directory. @pgwillia

README.md
pgwillia: ```rake migration:eraitem['spec/fixtures/migration/test-metadata/legacy-thesis-metadata/']``` gave me a thesis to test with.  Unfortunately I get the same error as before.
```
[root@hydranorthdev hydranorth]# bundle exec ruby bin/fix/fix-audit.rb 'thesis'
bin/fix/fix-audit.rb:73:in `find_objects': undefined method `length' for nil:NilClass (NoMethodError)
	from bin/fix/fix-audit.rb:111:in `fix_theses_related_dates'
	from bin/fix/fix-audit.rb:180:in `<main>'
```

weiweishi: try the attached foxml for the testing. This fix is to remove the theses
related date fields from non-thesis objects, so we need a non-thesis object
to test:


*Weiwei Shi*

2-10L Cameron Library
Phone:(780)492-7802
Fax: (780)248-1209
Email: weiwei.shi@ualberta.ca

weiweishi: [uuid_c5011e94-985a-4a6e-8fde-7554f0bd9813.xml.zip](https://github.com/ualbertalib/HydraNorth/files/294408/uuid_c5011e94-985a-4a6e-8fde-7554f0bd9813.xml.zip)

Try the foxml file inside.

pgwillia: @weiweishi even with the object you gave me this invocation (for both 'thesis' and 'reindex') seems to be expecting a second argument (see input = ARGV[1] on line 170. What should I try?

pgwillia:
@weiweishi in order to get this to work I used your sample 
```
#	spec/fixtures/migration/test-metadata/audit-fix-metadata/uuid_c5011e94-985a-4a6e-8fde-7554f0bd9813.xml
#	spec/fixtures/migration/test-metadata/collection/uuid_6d3695d8-74e1-44b2-a177-37635ad847b1.xml
#	spec/fixtures/migration/test-metadata/community/uuid_b13c373c-ea65-456d-8b53-07bc741a91ea.xml

```
and had to modify lib/tasks/migrate.rake to set the datesubmitted & dateaccepted (because they aren't set unless the item type is "Thesis").  I also had to make the following adjustments to get around errors. Does this seem right?
```
diff --git a/README.md b/README.md
index afd14b0..c3b05b5 100644
--- a/README.md
+++ b/README.md
@@ -131,14 +131,14 @@ Audit Fix
 * To fix data with triple files:
 ```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'triples','sample-sparql.txt'```
 * To remove irrelevant theses dates:
-```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'thesis'```
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'thesis' <pair tree prefix|id>```
 * To fix reindex:
-```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'reindex'```
+```RAILS_ENV = {{RAILS_ENV}} bundle exec ruby bin/fix/fix-audit.rb 'reindex' <pair tree prefix|id>```
 A set of rake tasks is also added for index jobs:
 rake hydranorth:solr:index[id]         Index a single object with ID
 rake hydranorth:solr:index_pairtree[input]   Index with a pairtree structure
 rake hydranorth:solr:update_generic_file_index  Index all Generic File objects
 
-/bin/fix/fix.rb is to update all the namespace uris
-/bin/fix/run.sh is to run script through all the pairtree combinations. 
+/bin/fix/fix.rb is to update all the namespace uris. Requires user to replace @server with the Fedora se
+/bin/fix/run.sh is to run script through all the pairtree combinations. Requires being run from the bin/
 
diff --git a/bin/fix/fix-audit.rb b/bin/fix/fix-audit.rb
index 20a779a..9855fb5 100644
--- a/bin/fix/fix-audit.rb
+++ b/bin/fix/fix-audit.rb
@@ -89,10 +89,10 @@ def find_objects(input)
           id = statement.object.relativize(statement.object.parent).to_s
          if id.include? "-"
             puts "Not a noid: " + id
-           notfixable += 1
+           @notfixable += 1
           else
            puts "An ERA Object: " + statement.object.to_s
-            fixable += 1
+            @fixable += 1
             objects << id
           end
         end
@@ -144,7 +144,7 @@ end
 
 def reindex_single(id)
   start = Time.now
-  ActiveFedora::Base.find(args[:id]).update_index
+  ActiveFedora::Base.find(id).update_index
   finish = Time.now
   used_time = finish - start
   puts "reindexed #{id} used #{used_time}"
diff --git a/bin/fix/fix.rb b/bin/fix/fix.rb
index 0d07b5b..6bd3c3c 100644
--- a/bin/fix/fix.rb
+++ b/bin/fix/fix.rb
@@ -8,7 +8,7 @@ instance = 'dev'
 
 $user = 'user'
 $password = 'password'
-$rest = 'http://@server:8983/fedora/rest'
+$rest = 'http://localhost:8983/fedora/rest'
 $originalroot = $rest
 
 $fixed = 0
```

weiweishi:
@pgwillia looks good to me. I think the migration script was changed to not use those date, but the initial migration used it so the data needed fixing.

Issue #1077, opened 2016-04-27
date_uploaded mapping


weiweishi:
According to the decision made on the Date Summit and following discussions with @sfarnel and @anayram , date_uploaded should use: info:fedora/fedora-system:def/model#createdDate to reflect the original date when the object first entered any of our systems. So current mapping need to be changed from dcterms:dateSubmitted to reflect this decision. 

This will also be reflected in the data change in #977

weiweishi:
closed by PR #1080

Issue #997, opened 2016-03-02
Fix namespace issue in ERA content and code


pbinkley:
As part of the audit process, actually fix the bad namespace identified in #986 in the content of ERA and in the code (if present).

Issue #941, opened 2016-01-26
Update rights language for theses


sfbetz:
We have a new release form for theses submitted to FGSR that will take effect at the end of January [exact date TBD]. Updated language is already incorporated into a test deploy of Thesis Deposit, and ready to go live at FGSR's go-ahead.

We would like to update all theses in ERA that would fall under any previous release forms (all existing theses, plus any theses to be added to ERA in the future that were submitted under the terms of the previous agreement) to **replace** their current rights statement with the following:

-Existing theses:
"This thesis is made available by the University of Alberta Libraries with permission of the copyright owner solely for the purpose of private, scholarly or scientific research. This thesis, or any portion thereof, may not otherwise be copied or reproduced without the written consent of the copyright owner, except to the extent permitted by Canadian copyright law."

I would like to have theses submitted to Thesis Deposit before that date (regardless of any embargo period), use the above statement.

Theses submitted to Thesis Deposit  AFTER we begin using the new release form, will have the following rights statement attached. 

-New theses:
"This thesis is made available by the University of Alberta Libraries with permission of the copyright owner solely for non-commercial purposes. This thesis, or any portion thereof, may not otherwise be copied or reproduced without the written consent of the copyright owner, except to the extent permitted by Canadian copyright law."

I'm not sure of the best way to do this... Would it make sense to:
1. update every existing thesis in ERA with the first statement
2. update every thesis that is currently under embargo in Thesis Deposit with the first statement, so they come into ERA with the correct language.

AND:
1. The second rights statement is currently deployed in development Thesis Deposit, and ready to go live when FGSR implements the new form, so all new theses would be submitted with the correct rights attached.

pbinkley:
After discussion with @sfbetz, we think this will work: 

- she will work out a change-over date with FGSR, along with final sign-off on the wording. On that date the new release of ThesisDeposit, containing the new wording, will be deployed. @henryzhang87 will be included in the choice of date so that he has warning about the deployment task.
- the timing of the appearance of the new wording on theses in HydraNorth is not a concern. We can start updating them any time. We'll do this in the rails console on Bournemouth (or with a rake task)
- We'll add the new wording to unembargoed theses as part of the migration task, as they move from ThesisDeposit to ERA: every thesis that moves will be given the new wording.

The changes will not be made until FGSR has signed off.

sfarnel:
Will update stylesheet for transform that occurs between ThesisDeposit and
ERA.

[...]

Issue #780, opened 2015-10-16
Rights statement cut off


sfarnel:
For some items the rights string is cut off.

E.g., https://hydranorth.library.ualberta.ca/files/wh246s128#.ViFwhkZSn9U

![righths](https://cloud.githubusercontent.com/assets/4697932/10554112/c2e8faae-741d-11e5-9660-981cf0fb50c8.jpg)

e.g., https://hydranorth.library.ualberta.ca/files/ff365650s#.ViFx_UZSn9U

pbinkley:
The rights statement seems to be drawn from the the LABEL attribute of the datastream in the foxml, instead of from the content of the datastream:

```<foxml:datastreamVersion ID="LICENSE.5" LABEL="Use of this product is restricted to current faculty,..." CREATED="2014-04-24T18:09:34.399Z" MIMETYPE="text/plain">```

The truncated form from the label was displayed in old ERA and you had to click the "download license" link to see the full text. This item fell through to the ```else``` at https://github.com/ualbertalib/HydraNorth/blob/master/lib/tasks/migration.rake#L421 . The downloading of the license file was only triggered by the presence of a filename in the LABEL attribute, handled by https://github.com/ualbertalib/HydraNorth/blob/master/lib/tasks/migration.rake#L370 .

We should import the full text for all items with this rights statement, as a data cleanup task (once we're past our current stability problems).

Full text:

> Use of this product is restricted to current faculty, staff, and students of the University. It is the responsibility of each user to ensure that he or she uses this product for individual, non-commercial educational or research purposes only, and does not systematically download or retain substantial portions of information.  Users may not reproduce or redistribute unprocessed/raw data portions of the data to any third party, or otherwise engage in the systematic retransmission or commercialization of the data.

pbinkley:
Resolve as part of audit.

sfarnel:
Captured in #977

PR #1087, opened 2016-05-06
update migration script to reflect changes in metadata and namespace …

weiweishi:
This PR is to reflect changes in metadata and namespace mapping in migration script:
* namespace changes reflects in #997 
* date_uploaded #1077 
* rights statement #941 and #780 
* not migrating date_accepted and date_submitted for non thesis objects.

### DIFF
--- a/lib/tasks/migration.rake
+++ b/lib/tasks/migration.rake
@@ -264,11 +264,9 @@ namespace :migration do
       unicorn = dc_version.xpath("ualid:unicorn", NS).text() if dc_version.xpath("ualid:unicorn", NS)
       degree_grantor = dc_version.xpath("marcrel:dgg", NS).text() if dc_version.xpath("marcrel:dgg", NS)
       dissertant = dc_version.xpath("marcrel:dis", NS).text() if dc_version.xpath("marcrel:dis", NS)
-      dissertant = creators.first if type == "Thesis" && (dissertant.nil? || dissertant.blank?)
-      end
+      dissertant = creators.first if (dissertant.nil? || dissertant.blank?)
       #calculated year_created based on date_created or date_accepted
-      if type == "Thesis"
-        year_created = date_accepted[/(\d\d\d\d)/,0] unless date_accepted.nil? || date_accepted.blank?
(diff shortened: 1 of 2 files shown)


lib/tasks/migration.rake
```diff
@@ -241,20 +252,20 @@ namespace :migration do
...
-      thesis_level = dc_version.xpath("ualterms:thesislevel", NS).text() if dc_version.xpath("ualterms:thesislevel", NS)
+      thesis_level = dc_version.xpath("ualthesis:thesislevel", NS).text() if dc_version.xpath("ualthesis:thesislevel", NS)
       alternative_titles = dc_version.xpath("dcterms:alternative/text()", NS).map(&:text) if dc_version.xpath("dcterms:alternative", NS)
-      proquest = dc_version.xpath("ualterms:proquest", NS).text() if dc_version.xpath("ualterms:proquest", NS)
-      unicorn = dc_version.xpath("ualterms:unicorn", NS).text() if dc_version.xpath("ualterms:unicorn", NS)
+      proquest = dc_version.xpath("ualid:proquest", NS).text() if dc_version.xpath("ualid:proquest", NS)
+      unicorn = dc_version.xpath("ualid:unicorn", NS).text() if dc_version.xpath("ualid:unicorn", NS)
       degree_grantor = dc_version.xpath("marcrel:dgg", NS).text() if dc_version.xpath("marcrel:dgg", NS)
       dissertant = dc_version.xpath("marcrel:dis", NS).text() if dc_version.xpath("marcrel:dis", NS)
       dissertant = creators.first if type == "Thesis" && (dissertant.nil? || dissertant.blank?)
```
pbinkley: We could remove the redundant check of type == "Thesis" - and is the case difference between this and line 246 (Thesis vs thesis) significant?

weiweishi: The check of type is to avoid date_accepted and date_submitted to be migrated for non-thesis objects, as they were bugs in old ERA and some regular ERA objects have that data as well. But the case should be "Thesis", pushing a change to the PR right now. @pbinkley

PR #1103, opened 2016-05-19
Add permitted attributes to user controller

pbinkley:


app/controllers/users_controller.rb
```diff
@@ -64,4 +64,11 @@ def set_saml
     end
   end
 
+  protected
+
+  def user_params
+    params.require(:user).permit(:avatar, :facebook_handle, :twitter_handle,
```
mbarnett: We don't use linked in, and we have a display_name attribute added. This should be
```ruby
  def user_params
    params.require(:user).permit(:display_name, :avatar, :facebook_handle, :twitter_handle,
                               :googleplus_handle, :remove_avatar, :orcid)
  end
```

for our purposes.

pbinkley:
Ah, should've caught that. I've committed the proper version.

PR #1116, opened 2016-05-19
Reduce Fedora traffic by removing download links from catalog

### LINKED ISSUE #1097: Add DOI links to Solr document
From https://github.com/ualbertalib/HydraNorth/issues/1091#issuecomment-219857973 : search results page rendering is slowed by the need to discover DOI urls for each item. Since they aren't in the Solr index, they have to be retrieved from Fedora. So: add them to metadata.rb, and confirm that this fixes the performance hit.

mbarnett:
also includes small tweaks to generic file page that may improve performance. Tests related
to doi links appearing on results pages are made pending. Closes #1097


pbinkley:
I'm not sure where it's being set, but in this branch doi_url is multivalued, while in ERA production it's single-valued:

ERA production:

```
irb(main):021:0> gf=GenericFile.find('jh343s338')
...
irb(main):019:0> gf.doi_url
=> "http://dx.doi.org/10.7939/DVN/10253"
```

Dev with this branch:

```
gf.doi_url=["http://dx.doi.org/10.7939/DVN/10253"]
=> ["http://dx.doi.org/10.7939/DVN/10253"]
irb(main):004:0> gf.doi_url
=> nil
irb(main):005:0> gf.doi_url="http://dx.doi.org/10.7939/DVN/10253"
ArgumentError: You attempted to set the property `doi_url' to a scalar value. However, this property is declared as being multivalued.
	from /usr/lib64/ruby/gems/2.1.0/gems/active-fedora-9.4.2/lib/active_fedora/attributes/property_builder.rb:20:in `doi_url='
...
```

Assigning an array to doi_url and saving does save the value into Fedora, but it doesn't get into Solr: gf.to_solr gives ```\"doi_url\":null,```.

mbarnett:
Interesting. I'll take a look. Should be a quick fix.

mbarnett:
Turns out, I've got a better way to do this.

PR #1136, opened 2016-05-25
set edit_form_class to reflect allowed terms for cstr et al and corre…

### LINKED ISSUE #1088: When deposit to collection with special id, id is not saved
e.g. "Computing science technical report ID" field isn't saving.

(spun off from #920)

Uncertain whether this bug applies to all saving (e.g. via the rails console)

pgwillia:
…ct title and description edit form param to pass array value fixes #1088

### DIFF
--- a/app/controllers/concerns/hydranorth/files_controller_behavior.rb
+++ b/app/controllers/concerns/hydranorth/files_controller_behavior.rb
@@ -11,12 +11,6 @@ module Hydranorth
     include Hydranorth::Breadcrumbs
 
 
-    included do
-      self.edit_form_class = Hydranorth::Forms::GenericFileEditForm
-      self.presenter_class = Hydranorth::GenericFilePresenter
-    end
-
-
     protected
 
     def actor
@@ -27,19 +21,6 @@ module Hydranorth
       attributes = params
     end
 
-    def presenter
-      if @generic_file[:resource_type].include? Sufia.config.special_types['cstr']
-        Hydranorth::CstrPresenter.new(@generic_file)
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['ser']
-        Hydranorth::SerPresenter.new(@generic_file)
-      elsif @generic_file[:remote_resource] == "dataverse"
-        Hydranorth::DataversePresenter.new(@generic_file)
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['thesis']
-        Hydranorth::ThesisPresenter.new(@generic_file)
-      else
-        Hydranorth::GenericFilePresenter.new(@generic_file)
-      end
-    end
 
     def edit_form
       find_collections_with_read_access
@@ -47,25 +28,6 @@ module Hydranorth
 
       @community_collections = collections_for_community(@generic_file.belongsToCommunity.first)
 
-      if @generic_file[:resource_type].include? Sufia.config.special_types['cstr']
-        Hydranorth::Forms::CstrEditForm.new(@generic_file)
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['ser']
-        Hydranorth::Forms::SerEditForm.new(@generic_file)
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['thesis']
-        Hydranorth::Forms::ThesisEditForm.new(@generic_file)
-      else
-        Hydranorth::Forms::GenericFileEditForm.new(@generic_file)
-      end
-    end
-
-    def update_metadata
-      if @generic_file[:resource_type].include? Sufia.config.special_types['cstr']
-        self.edit_form_class = Hydranorth::Forms::CstrEditForm
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['ser']
-        self.edit_form_class = Hydranorth::Forms::SerEditForm
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['thesis']
-        self.edit_form_class = Hydranorth::Forms::ThesisEditForm
-      end
       super
     end
 
--- a/app/controllers/concerns/hydranorth/files_controller_behavior.rb
+++ b/app/controllers/concerns/hydranorth/files_controller_behavior.rb
@@ -58,6 +58,16 @@ module Hydranorth
       end
     end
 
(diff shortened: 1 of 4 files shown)


pgwillia:
Failures:

  1) collection create collection should be able to create a collection
     Failure/Error: collection_id = Collection.where(title: 'TESTTEST').first.id
     NoMethodError:
       undefined method `id' for nil:NilClass
     # ./spec/features/collection/collection_spec.rb:368:in `block (3 levels) in <top (required)>'

pgwillia:
app/views/records/edit_fields/_description.html.erb && app/views/records/edit_fields/_title.html.erb are used by both genericfile & collection edit views.

pgwillia:
Returning to a bit of a hack to make this work but should be ready to go.

app/controllers/concerns/hydranorth/files_controller_behavior.rb
```diff
@@ -58,6 +58,16 @@ def edit_form
...
+    def update_metadata
+      if @generic_file[:resource_type].include? Sufia.config.special_types['cstr']
+        self.edit_form_class = Hydranorth::Forms::CstrEditForm
+      elsif @generic_file[:resource_type].include? Sufia.config.special_types['ser']
+        self.edit_form_class = Hydranorth::Forms::SerEditForm
+      elsif @generic_file[:resource_type].include? Sufia.config.special_types['thesis']
+        self.edit_form_class = Hydranorth::Forms::ThesisEditForm
+      end
+      super
+    end
```
mbarnett: Can we DRY this up a bit? The presenter and edit_form methods are repeating the same pattern as well. 

I also think there's probably not a lot of value in trying to deal with controller logic in a module, when we're not expecting downstream users to mix this in. We don't have to clean that all up, but we can start down that path.

I'm thinking  delete ```update_metadata```, ```presenter```, and ```edit_form``` (or at least trim ```edit_form``` back to just https://github.com/ualbertalib/HydraNorth/blob/70002cc9f611c6834d062c1076e1eacf860905aa/app/controllers/concerns/hydranorth/files_controller_behavior.rb#L45-L48 if that's not already happening in its ```super```.

Then in ```GenericFilesController``` clean up the repeated checks in all three methods into something like ```cstr_resource?```, ```ser_resource?```, ```dataverse_resource?```, and ```thesis_resource?```.

*Then*, in , you should be able to just do something like this:

```ruby
def presenter_class
    super unless @generic_file
    return Hydranorth::CstrPresenter.new(@generic_file) if cstr_resource?
    # ... etc for every resource type, return Hydranorth::GenericFilePresenter.new(@generic_file) as a last result
end

def edit_form_class
  super unless @generic_file
  return Hydranorth::Forms::CstrEditForm.new(@generic_file) if cstr_resource?
  # etc
end
```

This way we don't have to keep overriding every client method of ```presenter_class``` and ```edit_form_class``` with the same logic. We're just patching the actual methods themselves.

WARNING: This is totally untested on my part so there may be some hidden difficulty I'm not anticipating. But I think it's worth starting down the path of disentangling this code now.

pgwillia: @mbarnett good suggestions.  I'm trying this now.  Might be some bleed over into the batch controller.

pgwillia:
@mbarnett I think this is ready for another look.

Finished in 28 minutes 43 seconds (files took 13.29 seconds to load)
404 examples, 0 failures, 3 pending

app/controllers/concerns/hydranorth/files_controller_behavior.rb
```diff
@@ -27,18 +41,9 @@ def attributes
...
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['ser']
-        Hydranorth::SerPresenter.new(@generic_file)
-      elsif @generic_file[:remote_resource] == "dataverse"
-        Hydranorth::DataversePresenter.new(@generic_file)
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['thesis']
-        Hydranorth::ThesisPresenter.new(@generic_file)
-      else
-        Hydranorth::GenericFilePresenter.new(@generic_file)
-      end
+      presenter_class.new(@generic_file)
```
mbarnett: This is how it's already defined in sufia, so we don't have to override presenter at all anymore https://github.com/projecthydra/sufia/blob/6.x-stable/app/controllers/concerns/sufia/files_controller_behavior.rb#L171

app/controllers/concerns/hydranorth/files_controller_behavior.rb
```diff
@@ -47,18 +52,9 @@ def edit_form
...
-      if @generic_file[:resource_type].include? Sufia.config.special_types['cstr']
-        Hydranorth::Forms::CstrEditForm.new(@generic_file)
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['ser']
-        Hydranorth::Forms::SerEditForm.new(@generic_file)
-      elsif @generic_file[:resource_type].include? Sufia.config.special_types['thesis']
-        Hydranorth::Forms::ThesisEditForm.new(@generic_file)
-      else
-        Hydranorth::Forms::GenericFileEditForm.new(@generic_file)
-      end
+      edit_form_class.new(@generic_file)
```
mbarnett: ditto this https://github.com/projecthydra/sufia/blob/6.x-stable/app/controllers/concerns/sufia/files_controller_behavior.rb#L179

app/controllers/concerns/hydranorth/files_controller_behavior.rb
```diff
@@ -12,8 +12,22 @@ module FilesControllerBehavior
 
 
     included do
-      self.edit_form_class = Hydranorth::Forms::GenericFileEditForm
-      self.presenter_class = Hydranorth::GenericFilePresenter
+      def presenter_class
+        super unless @generic_file
+        return Hydranorth::CstrPresenter if cstr_resource?
```
mbarnett: Since this depends on the methods in FilesController anyways, why not move it over there? It looks like that's the only place this is mixed in anyways so it shouldn't impact anything else.

pgwillia:
@mbarnett how about now?

mbarnett:
👍

PR #1164, opened 2016-06-01
1075 restricted communities

criedlberger:


spec/controllers/selects_collections_spec.rb
```diff
@@ -58,7 +58,7 @@ class SelectsCollectionsController < ApplicationController
 
         it "should return only public and read access (edit access implies read) communities" do
```
pgwillia: This comment isn't true anymore

app/models/ability.rb
```diff
@@ -18,7 +18,7 @@ def custom_permissions
       end
       cannot :create, ::Collection 
       cannot :destroy, ::Collection unless admin?
-      cannot :manage, ::Collection do |obj|
+      cannot [:edit, :update], ::Collection do |obj|
```
pgwillia: manage is ANY action.  Do we have to worry about create, destroy or any other actions here?

pgwillia:
Other than my two questions this looks good to go!

app/models/ability.rb
pgwillia: Nm.  Create and destroy are handled in the two lines above!

PR #1165, opened 2016-06-01
Sprint40

pbinkley:
These two commits were on the Sprint40 branch and didn't get merged into master. If @mbarnett can confirm that they haven't been superceded by other work, they can be merged.


mbarnett:
The code in my two commits there is already in master (I just double-checked).

I wouldn't merge this as-is, because it might clobber more recent changes.

PR #1173, opened 2016-06-06
704 html markup

criedlberger:


pgwillia:
@criedlberger could you please remove the 155 commit from this branch.

PR #1179, opened 2016-06-07
when the document is instutionally protected show a default thumbnail…

### LINKED ISSUE #1159: Front page contains http download of sign-in form
The front page of ERA in production currently attempts to download http://era.library.ualberta.ca/users/sign_in. It gets redirected to https, but it causes the page to be flagged as insecure in Firefox's location bar. The request seems to be caused by an attempt to load the thumbnail of a ccid-protected item in the "Recent Documents" box (currently https://era.library.ualberta.ca/downloads/3j333490f?file=thumbnail)

Just for context: This problem of ccid-protected thumbnails shows up elsewhere too. If you aren't logged in and you browse a collection that contains a ccid-protected item, you get that bar at the top of the results telling you you need to log in to see "this item". If you log in, you're taken to a thumbnail image. I assume this is what triggered the message, and Sufia assumes that it is what you were trying to see when you logged in.

pgwillia:
… fixes #1159

### DIFF
--- a/app/helpers/sufia_helper.rb
+++ b/app/helpers/sufia_helper.rb
@@ -2,4 +2,25 @@ module SufiaHelper
   include ::BlacklightHelper
   include Sufia::BlacklightOverride
   include Sufia::SufiaHelperBehavior
+
+  # override SufiaHelperBehavior method for institutional_access? condition
+  def sufia_thumbnail_tag(document, options)
+    # collection
+    if document.collection?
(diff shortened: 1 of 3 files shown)


pgwillia:
Note that this won't change the stored location so some users may still experience the redirect to thumbnail download on sign in.  This also doesn't address the redirect to http from the image tag.

weiweishi:
Had one failed spec solved by PR #1182

PR #1189, opened 2016-06-08
redefining spec means `rake spec` runs tests twice fixes #1180

### LINKED ISSUE #1180: Test suite needs attention
We have reports that all the specs are running twice, or that they are extremely slow. Let's have a look at whether something has changed that can be fixed.

pgwillia:
Doesn't address @weiweishi slow running tests though...


weiweishi:
Tests are running much faster (24 mins) after given my VM 8 GB of RAM.

Issue #528, opened 2015-07-20
Service Migration: handles


pbinkley:
ERA and DSpace handles must be updated to resolve to new location.

weiweishi:
closed with PR #651

weiweishi:
@pbinkley Does it mean we need to update the handle servers with the new location?

pbinkley:
Yes - currently the handle server resolves to the old ERA url and then
HydraNorth redirects to the new url. We ought to have the handle resolve
directly to the new url. This isn't urgent, since it's working for users
now, but we ought to clean it up. We can do this after all issues from the
migration audit are resolved, since they might affect a handful of items
that need updates for their handles.

Peter Binkley, Ph.D., MLIS
Digital Initiatives Technology Librarian
Information Technology Services
peter.binkley@ualberta.ca

2-10K Cameron Library
University of Alberta
Edmonton, Alberta
Canada T6G 2J8

phone 780-492-3743
fax 780-492-9243

PR #1190, opened 2016-06-09
Migration handle update

weiweishi:
* allow migration process to decode html encodings such as $amp;
* update the script so that it will update handles in production to redirect to ERA
* generate noid report at the end of the migration process.


lib/tasks/migration.rake
```diff
@@ -702,9 +714,38 @@ namespace :migration do
       puts "Summary: Save file time: #{save_time}"
       puts "Summary: Add to Collection time: #{collection_time}"
       puts "Summary: Verification time: #{verify_time}"
-
+      noid_report.close
   end
 
+    def update_handle(handle,noid)
+      link = "https://era.library.ualberta.ca/files/#{noid}"
+      handle.slice!("http://hdl.handle.net/")
```
pgwillia: Is it future proof/safe to have the link and handle target hard coded like this?

lib/tasks/migration.rake
```diff
@@ -702,9 +714,38 @@ namespace :migration do
...
+      MigrationLogger.info "Deleting existing handle #{handle}"
+      puts handle
+      delete_cmd = "./bin/handle/bin/hdl-delete 0.NA/10402 bin/handle/lib/admpriv.bin #{handle}"
```
pgwillia: How will the handle cli be installed in this path? Can we guarantee it will be? I don't see it in the project's bin directory currently or as part of this PR.

config/initializers/sufia.rb
```diff
@@ -208,7 +208,8 @@
      "2014-06" => "2014-06",
      "2014-11" => "2014-11",
      "2015-06" => "2015-06",
-     "2015-11" => "2015-11"
+     "2015-11" => "2015-11",
+     "2016-06" => "2016-06"
```
pgwillia: Could these values be read from a database?  Seems unsustainable to continue to add to initializer. Also not related to the handle update, right?

lib/tasks/migration.rake
```diff
@@ -232,7 +234,8 @@ namespace :migration do
       creators = dc_version.xpath("dcterms:creator/text()", NS).map(&:text) if dc_version.xpath("dcterms:creator", NS)
       contributors = dc_version.xpath("dcterms:contributor/text()", NS).map(&:text) if dc_version.xpath("dcterms:contributor",NS)
       subjects = dc_version.xpath("dcterms:subject/text()",NS).map(&:text)
-      description = dc_version.xpath("dcterms:description",NS).text.gsub(/"/, '\"').gsub(/\n/,' ').gsub(/\t/,' ')
+      description = dc_version.xpath("dcterms:description",NS).text.gsub(/\n/,' ').gsub(/\t/,' ')
+      description = HTMLEntities.new.decode description
```
pgwillia: Is there a spec fixture that illustrates the before and after of the HTMLEntities decode group of changes?

lib/tasks/migration.rake
```diff
@@ -616,6 +621,12 @@ namespace :migration do
 
       MigrationLogger.info "Generic File saved id:#{@generic_file.id}"
       MigrationLogger.info "Generic File created id:#{@generic_file.id}"
+      puts "Updating handle #{fedora3handle} for migrated file #{@generic_file.id}"
+      MigrationLogger.info "Updating handle #{fedora3handle} for migrated file #{@generic_file.id}"
+      #This is to call a command line script to update the handle to redirect to ERA/noid
+      #Please follow the instruction here https://github.com/ualbertalib/era/tree/handle/handle to configure the scripts to work with the migration
+ 
+      update_handle(fedora3handle, @generic_file.id) if ENV['RAILS_ENV'] == 'production'
```
pgwillia: Suggestions for testing in another environment? Not sure if I feel comfortable trusting that this will just work.

pgwillia:
Is this PR related to any open issues?

config/initializers/sufia.rb
weiweishi: @pgwillia yes, this is not related to handle update, or migration. This is for the edit form to have the value presented (which is only available for admin). Maybe we want to revisit how the values are populated on the edit form.

pgwillia: Then it probably doesn't belong in this pull request.  Or at least not the same commit.

weiweishi: Does it make sense to create a PR with this one line change? This has to be in, in order for admins to edit the newly migrated objects.

lib/tasks/migration.rake
weiweishi: I will add a spec.

lib/tasks/migration.rake
weiweishi: You can just try this script on your desktop for testing. It doesn't need rails environment. It's a standalone script and we just called it as a shell command here. We removed a previous commit that added the script into Hydranorth as it contains authentication information and it's currently in a private repo. Please follow the instruction to test it.

lib/tasks/migration.rake
weiweishi: I will update the era link, and put the handle link into a constant in the script.  That value is very unlikely to be changed in the future.

lib/tasks/migration.rake
weiweishi: @pgwillia  It's been pulled out from the PR as it contains private authentication information. It's now installed on bournemouth. It probably makes more sense to use Ansible playbook to handle the installation. I will check with @henryzhang87

config/initializers/sufia.rb
pgwillia: I don't feel right about hiding unrelated changes in biggish commits.  If you don't feel it warrants a new PR, at least make it a separate commit so we can easily see the history.

weiweishi: Okay. I will do a separate commit if that's acceptable to you @pgwillia working on this now.

weiweishi:
We didn't create a issue for the thesis migration, but it sort of belongs here:
https://github.com/ualbertalib/HydraNorth/issues/528
Same script and strategy can be used to solve this issue.

Issue #962, opened 2016-02-10
Redesign Collections landing page


sfbetz:
As a user I want to be able to browse sets of items grouped by collections so I can find related items.

As a user, I want options to sort and refine sets of items, so I can easily find what I’m looking for.

Specifications here: 
https://docs.google.com/presentation/d/195k-GIB4C2W7KFrqna6XMki8IGxvWaZx_HGBXZSYzh8/edit?usp=sharing

Same strategy as Community Page #961, with exceptions:
-no logo or icon required.
-different labels

sfbetz:
part of epic #840

weiweishi:
PR #1029

sfbetz:
Some comments and bugs:

Bug in facets - clicking on "more" in facets doesn't display list when logged in, Public view not allowing faceting
Public view of landing page doesn't include paging and sorting options
Pagination not displaying on either logged in or public view
Don't need description button - can just display description text
Logged in view search results are squeezed into middle of the page.

pbinkley:
Closes #786

PR #1193, opened 2016-06-15
Refactor collections to allow nesting

### LINKED ISSUE #909: Collection refactoring
As admin, I want the ability to nest collections into communities so I can easily organize official collections and communities. 

-when I create a new collection, I want to be able to choose a community to add it to.
-from an existing community, I want to be able to add collections to the community
-from an existing collection, I want to be able to add that collection to a community
-from the dashboard, "my communities and collections" I want to be able to select a collection and add it to a community.

@weiweishi do you have documentation as to how we are doing this without using the interface?

mbarnett:
Touches more or less everything. Closes #909.


app/controllers/concerns/hydranorth/collections/selects_collections.rb
```diff
@@ -1,5 +1,6 @@
 module Hydranorth::Collections::SelectsCollections
   extend ActiveSupport::Concern
+  # shouldn't this be Sufia::Catalog? We're all over the place with includes
```
weiweishi: This was copied directly from Hydra:Collections to overwrite. But I think you are right, should've used Sufia::Catalog instead.

weiweishi:
@mbarnett When removing children collections from parent collection from the action menu "Remove from Collection"
screenshot - 16-06-23 - 03 03 32 pm

It gives the following error message:
screenshot - 16-06-23 - 03 04 21 pm

weiweishi:
After merging with latest master, and with the collection page update, the action menu item "Remove from collection" is no longer available.

mbarnett:
Did a different ticket remove the "Remove from collection" menu item? Or is that a bug?

weiweishi:
@mbarnett It was a bug fixed in #962 but then that feature branch has never merged into master. However, "Remove from Collection" doesn't work properly. This is out of scope for this PR though. I will open another issue and ask @leahvanderjagt @sfbetz to prioritize.

app/models/concerns/hydranorth/collection_behavior.rb
```diff
@@ -12,7 +12,7 @@ module CollectionBehavior
     include Hydranorth::Collections::Logo
 
     included do
-      before_save :remove_self_from_members, :update_permissions, :check_logo_size
+      before_save :check_logo_size, :remove_self_from_members, :update_permissions
```
weiweishi: Does the order of these functions run before_save matter?

weiweishi:
@mbarnett Have one failed spec: 
  1) collection community landing page as user should have following features
     Failure/Error: within("#facets") do
     Capybara::ElementNotFound:
       Unable to find css "#facets"
     # ./spec/features/collection/collection_spec.rb:77:in `block (3 levels) in <top (required)>'

mbarnett:
Thanks, I'll take a look

weiweishi:
@mbarnett This seems to be caused by no objects are displayed at the community level. And it is because of the change here: https://github.com/ualbertalib/HydraNorth/blob/b2c52562ee4232556801e1a30502fe7060cdfa7f/app/models/search_builder.rb#L8

mbarnett:
It should be possible to fix that. The question is, do we want to, or should we fix the test, instead? Right now it's implementing the behaviour of not showing generic files in top-level Communities, which I think is the long-term desire for functionality with Communities.

weiweishi:
Double checked with @sfbetz. As there are many objects currently in ERA that belong only to communities but not collections, we will need to still be able to display them. The long-term goal is to clear those objects and move them into collections. We will make a note on #1205 that we will revert back to what you have right now once the data cleaning is done. Meanwhile, could you please comment out this, and use a logic that will show the objects on the community level? Thanks! @mbarnett

mbarnett:
Sounds good, will commit that this afternoon

mbarnett:
Pushed fixes for files not showing on Community pages, sorting in the collection pop-up, and when changing the pagination on a dashboard tab no longer sends you back to the files tab. 

I'm still looking into sorting the actual list of files and collections on the dashboard programmatically. The trick is that title is multi-valued in the Solr index (even though we're not using it that way) so Solr refuses to sort by it. We might need to re-index to fix this, but I still have a couple more ideas to try.

weiweishi:
all spec passes except for the one identified in PR #1206. 

  1) collection collection with a member file should not have edit and delete options
     Failure/Error: expect(page).to have_content("Download")
       expected to find text "Download" in "Login education & research archive ERA Search ERA Login home browse search upload your research contact Search ERA Advanced Search Browse communities, subjects, or item types Test Collection 1 entry found Sort by Relevance ▼ Relevance ▼ Date (newest first) Date (oldest first) New items Date modified (newest first) Date modified (oldest first) Number of results to display per page 10 per page 10 per page 20 per page 50 per page 100 per page Items in this Collection Search Collection Test Collection Go Test Item Title: Test Item Creator: Subjects: Quick Links Login Browse Search Share Your Research Policies About ERA Policies Technology and Partnerships Get Help Contact Us How to Deposit © University of Alberta Libraries | University of Alberta, Edmonton, AB, Canada T6G 2R3 | Web Site Privacy Policy"
     # ./spec/features/collection/collection_spec.rb:367:in `block (3 levels) in <top (required)>'

weiweishi:
@mbarnett I will merge this PR, and if you figured out sorting the list on the document list, it should probably associated with #1205?

PR #1197, opened 2016-06-17
move rest-client to inside the rake task function

weiweishi:


mbarnett:
You need to remove it on line 2, too 😉  https://github.com/ualbertalib/HydraNorth/blob/d2611758ca592ce6a6007a5d6e33ec085800be09/lib/tasks/hydranorth-solr.rake#L2

weiweishi:
apparently I'm really bad at multitasking 😉 . Done now. Thanks @mbarnett

mbarnett:
Verified that this works (locally) by temporarily commenting ```rest-client``` out of the Gemfile, uninstalling the gem, and then running ```db:migrate``` successfully.

LGTM. Merging.

Issue #1195, opened 2016-06-16
Sitemap error due to broken items


pbinkley:
add error-handling to sitemap generation to log broken items for manual repair

pgwillia:
@weiweishi or @anayram is it easy to get the metadata of an object (or a small group of them) that exhibits these symptoms?

pgwillia:
To reproduce got uuid:7a385f26-6387-4f4e-9ba6-152044731c04.xml migrated with v1.2.1 output was
```
ERROR [line: 34] With input '"Smith, John B. \"Web-Based Systems and Instruction.\" Web. &lt; http://www.cs.unc.edu/Research/jbsAr': Invalid token "\"Smith," (found "\"Smith,"), production = :RDFLiteral
ERROR [line: 34] Unexpected (found "http:"(PNAME_NS)), production = "."
ERROR [line: 35] Unexpected (found "A"), production = ")"
ERROR [line: 35] undefined prefix "http"
ERROR [line: 35] With input '//www.cs.unc.edu/Research/jbsArchive/docs/AutoGeneratedSystems/ &gt; Accessed 31 March, 2015.
	Smith,': Invalid token "//www.cs.unc.edu/Research/jbsArchive/docs/AutoGeneratedSystems/" (found "//www.cs.unc.edu/Research/jbsArchive/docs/AutoGeneratedSystems/"), production = :predicateObjectList
ERROR [line: 36] With input 'Smith, John B. and Catherine F. Smith. \"ChicoryLane Farm.\" Website. &lt; http://www.chicorylane.com': Invalid token "Smith," (found "Smith,"), production = :_turtleDoc_1
ERROR [line: 34] With input '"Smith, John B. \"Web-Based Systems and Instruction.\" Web. &lt; http://www.cs.unc.edu/Research/jbsAr': Invalid token "\"Smith," (found "\"Smith,"), production = :RDFLiteral
ERROR [line: 34] Unexpected (found "http:"(PNAME_NS)), production = "."
ERROR [line: 35] Unexpected (found "A"), production = ")"
ERROR [line: 35] undefined prefix "http"
ERROR [line: 35] With input '//www.cs.unc.edu/Research/jbsArchive/docs/AutoGeneratedSystems/ &gt; Accessed 31 March, 2015.
[...]

PR #1199, opened 2016-06-21
add error handling to sitemap solves #1195

pgwillia:


weiweishi:
all spec passes with the fix in PR #1206

PR #1204, opened 2016-06-23
Migration handle update

weiweishi:
-     allow migration process to decode html encodings such as & amp;
-     update the script so that it will update handles in production to redirect to ERA
-     generate noid report at the end of the migration process.
-     added the new graduation date to allow editing in UI
This is related #528 as it can be solution with same script and strategy if needed.


weiweishi:
@pgwillia I messed up with the commit history in the other PR #1190 So I've created this new one. Please let me know if I have addressed all the concerns you have. Thanks!

pgwillia:
@weiweishi did you follow up with @henryzhang87 about installing the handle cli from the playbook? This is the only part I'm still concerned about. Shall I merge as is and create a technical maintenance ticket to ensure that this will fail gracefully? I don't think this even reports failure at the moment.

pgwillia:
Otherwise I think this looks good :+1:

weiweishi:
@pbinkley @mbarnett this is ready now.

Issue #1071, opened 2016-04-22
File download link change


pbinkley:
Create a route for file download: instead of ```https://era.library.ualberta.ca/downloads/j098zd881```, use the file label (i.e. the original file name) and place it under ```/files/``` instead of ```/downloads/```, to produce ```https://era.library.ualberta.ca/files/j098zd881/120715_parliamentnow.pdf```

This meets Google's and LAC's requirements, for the downloadable file to be in the same "directory" as the landing page for the item, and for the download link to end in ".pdf" for thesis pdfs.

The old url should issue a permanent redirect to the new one: it will have to do a lookup to determine the file label, in the way we do with old ERA uuids: #651

Closes #848, #849

pbinkley:
This structure is to be maintained alongside the ```/downloads/noid.pdf``` version.

PR #1213, opened 2016-07-06
fix to #1071 and #1212 to use filename when file.label not present

weiweishi:


weiweishi:
@pbinkley here is my fix to the issue when file.label is nil.

PR #1216, opened 2016-07-15
move admin_target_collection logic to Hydranorth::Collections::Select…

### LINKED ISSUE #1207: Only 10 Collections showing in pop-up when nesting collections
That pop-up should be an unwieldy list of all of the collections in the repository, but for some reason on Plano, when signed in as dittest@ualberta.ca, it was only listing 10 collections.

This is blocking use of the collection nesting functionality.

pgwillia:
…sCollections fixes #1207 

* Hydranorth::Collections::SelectsCollections uses the ::CollectionSearchBuilder class which changes the number of rows returned to 1000
  * @target_collections is similar to the @user_collections but different.  @user_collections is used on the communities browse pages while @target_collections will be used in the collection selection dialog.
* also use Solrizer to generate solr_name


pgwillia:
@pbinkley @mbarnett here's another bug fix.

mbarnett:
I'm not sure I like folding the AdminNestingTargets code into SelectsCollections? That module was already kind of an all-slicing-all-dicing God Module servicing a bunch of different use cases, and adding the Admin stuff into it just kind of leads it further down that path.

I think we should look at teasing apart SelectsCollections, rather than doing this. Either way I don't think we can rush this in today.

app/controllers/catalog_controller.rb
```diff
@@ -16,7 +15,6 @@ class CatalogController < ApplicationController
   skip_before_filter :default_html_head
 
   def index
```
mbarnett: if this is just going to be a call to super, it should be removed entirely.

pgwillia:
I agree that this doesn't need to be merged today. Thanks for looking at it.

The bug with the collection selection modal showing only 10 collections was fixed by creating CollectionSearchBuilder and forcing Sufia to use it in Hydranorth::Collections::SelectsCollections with the collections_search_builder_class definition. The admin targets and the user collections is a similar problem -- the user is selecting collections, I don't see why they can't be solved in the same module.

mbarnett:
> The admin targets and the user collections is a similar problem -- the user is selecting collections, I don't see why they can't be solved in the same module.

Well, I'd like to refactor it into a few different modules, that are a client of the same fix, really. The problem as I see it is that moving the population of ```@target_collections``` into ```find_collections``` results in that being available almost everywhere in the application (well, at the very least, in the file, collection, community, catalog, and dashboard controllers, and anywhere else that calls anything in SelectCollections), whereas previously it was only available in the Catalog and dashboard controllers.

SelectCollections is like this dumping ground of quasi-global variables which it seems like we keep throwing more code into, even though a lot of the things that include it only need a small portion of the code. If we can split it up in such a way that we share things between modules but each controller only includes what it actually needs, I think it would be a bit more tractable for bug fixing. We're getting a bit big ball of mud-dy with all of this shared state between different controllers.

mbarnett:
Oh, batch controller uses it too. Probably worth sketching out what the different ways everything uses different parts of this are, and trying to find some natural seems to break this up along.

pgwillia:
I can see the 'big ball of mud-dy'ness. I'm guilty of finding something that works and not thinking about it any further because the Blacklight/Sufia tangles are already pretty confusing to me. I'm pretty sure that's why SelectsCollection is as unreadable as it currently is.

Using SelectsCollection in the dashboard controller makes sense (that's where you select collections).  Not sure about the catalog controller (maybe because admin can select from there?).  The files controller and batch controller also makes sense because in the edit we have those drop downs to add to collection. I think it'll be difficult to find natural seams to break this up with because I **think** it's being used appropriately most of the time.  We **might** be able to remove SelectsCollection from some of the controllers which would be a win.

pgwillia:
:+1:  sounds good to me!

PR #1229, opened 2016-08-08
1205 Implement sortable title indexing & resolve issue w/ not all collections being targetable

mbarnett:
@pgwillia tagging you with this since I'd like to get your input on the refactoring I've done

Unfortunately I didn't squash prior to merging in Tricia's work, and it's not worth the mess of linearizing that.


app/controllers/concerns/hydranorth/batch_controller_behavior.rb
```diff
@@ -3,21 +3,19 @@ module BatchControllerBehavior
...
-
       class_attribute :edit_form_class, :cstr_edit_form_class, :ser_edit_form_class
       self.edit_form_class = Hydranorth::Forms::BatchEditForm
       self.cstr_edit_form_class = Hydranorth::Forms::CstrBatchEditForm
       self.ser_edit_form_class = Hydranorth::Forms::SerBatchEditForm
     end
 
     def edit
+      @user_communities = find_communities(access_levels[:edit])
+      @user_collections = find_collections(access_levels[:edit])
```
pgwillia: Why use find_communities instead of find_communities_with_edit_access? Perhaps now we could test that a call to the edit method assigns @user_communities & @user_collections?

pgwillia: To answer my own question -- because find_collections_with_edit_access doesn't exist anymore.

pgwillia: Can we do the same to find_communities_with_edit_access?

pgwillia:
This is going to be so much better! Admin, Communities and Collection selection separated and sorting by title.

I think that with the refactor, that it should be easier to test the different selections scenarios. I'd like to see some specs added for the admin case.

For the title sorting.  Is there a unit or functional test we can add to ensure we don't end up with a regression?

app/controllers/concerns/hydranorth/collections/admin_collection_selection.rb
```diff
@@ -0,0 +1,10 @@
+module Hydranorth::Collections::AdminCollectionSelection
+  extend ActiveSupport::Concern
+  include Hydranorth::Collections::BaseQuery
+
+  def admin_target_collections
```
pgwillia: Would like to see a test for this.

app/controllers/concerns/hydranorth/collections/base_query.rb
```diff
@@ -0,0 +1,27 @@
...
+    ::CollectionSearchBuilder
+  end
+
+  def access_levels
+    { read: [:read, :edit], edit: [:edit] }
+  end
+
+  def perform_collection_query(query_string, access_level)
+    # need to check for _tesim and _bsi in solr query because ActiveFedora does not allow false to be passed
+    # need to know the user if there is an access level applied otherwise we are just doing public collections
```
pgwillia: This comment refers to the logic that would be in the query_string. Makes more sense if it's close to  app/controllers/concerns/hydranorth/collections/collection_selection.rb#6 and similar.

app/controllers/concerns/hydranorth/batch_controller_behavior.rb
mbarnett: Whoops, yeah.

The goal is to get rid of anywhere "hidden" that magically assigns to variables like @user_collection that get used in the view, and make it all explicit in the controller so we can see where this stuff is coming from (and prevent bugs where we munge it for one use-case and then it gets silently shared with some other controller/view accidentally).

mbarnett:
Good call on the added testing. I'll start working on that.

app/controllers/concerns/hydranorth/collections/base_query.rb
mbarnett: 👍

pgwillia:
I was looking at the title sort indexing more closely.  I ran ```rake migration:era_collection_community['/vagrant/migration/communities/']``` and ``` rake migration:era_collection_community['/vagrant/migration/collections/']``` (this is the full compliment of communities and collections from ERA), ```rake hydranorth:reindex_collections``` and created a Collection in the Web UI. In the solr index I don't see any fields with ```sortable_title_si``` at any stage in the process. The collections are all present and I don't see anything glaring in the logs. What am I missing?

mbarnett:
Can you try throwing a breakpoint on https://github.com/ualbertalib/HydraNorth/pull/1229/files#diff-b6e5c69d5ef46418b2d70629b9b984e9R6 and seeing if it hits and what object and title evaluate to? I can't imagine why that's not working (which probably means my imagination is limited).

pgwillia:
Wait a tick.  To answer my own question, of course I won't see anything.  It's not a stored field! Sorting works fine :+1: 
```
http://hydranorthdev.library.ualberta.ca:8983/solr/development/select?q=active_fedora_model_ssi:Collection&wt=json&indent=true&fl=title_tesim&rows=1600&sort=sortable_title_si%20asc
```
```
{
  "responseHeader":{
    "status":0,
    "QTime":4,
    "params":{
      "q":"active_fedora_model_ssi:Collection",
      "indent":"true",
      "fl":"title_tesim",
      "sort":"sortable_title_si asc",
      "wt":"json"}},
  "response":{"numFound":388,"start":0,"docs":[
      {
        "title_tesim":["2010 2nd Annual Celebration of Research and Innovation"]},
      {
        "title_tesim":["2012 Campus Data Summit"]},
      {
        "title_tesim":["2013 Campus Data Summit"]},
      {
        "title_tesim":["50th Anniversary Celebration"]},
      {
        "title_tesim":["Aboriginals"]},
      {
        "title_tesim":["ACCES Research Group"]},
      {
        "title_tesim":["ACRRE Research Notes"]},
      {
        "title_tesim":["Ad $ Summary"]},
      {
        "title_tesim":["Alberta Centre for Active Living"]},
      {
        "title_tesim":["Alberta Centre for Reclamation and Restoration Ecology (ACRRE)"]}]
  },
  "facet_counts":{
    "facet_queries":{},
    "facet_fields":{
      "active_fedora_model_ssi":[
        "Collection",388],
      "object_type_si":[]},
    "facet_dates":{},
    "facet_ranges":{},
    "facet_intervals":{}},
  "spellcheck":{
    "suggestions":[
      "correctlySpelled",true]}}
```

mbarnett:
Right, right, that makes sense.

*cancels exorcist appointment*

*drinks more :coffee: to jump-start brain*

pgwillia:
We just discussed using
```<dynamicField name="*_ssort" type="alphaSort" stored="false" indexed="true" multiValued="false"/>``` which uses [ICUFoldingFilterFactory](https://wiki.apache.org/solr/AnalyzersTokenizersTokenFilters#solr.ICUFoldingFilterFactory) instead of using Ruby's downcase.

**Comment by mbarnett** [comment]
It looks like we will need to extend [Solrizer's descriptors](https://github.com/projecthydra/solrizer/blob/d6b2e2d65041a0d204b4541556c72298b227fca4/lib/solrizer/default_descriptors.rb) to accomplish that, so rather than drag this out I'm going to throw a separate ticket in the backlog to make title sorting unicode-clean.

**Comment by weiweishi** [comment]
Tried on plano:
http://plano.library.ualberta.ca:8983/solr/development/select?q=has_model_ssim%3ACollection&sort=sortable_title_si+asc&wt=json&indent=true after reindex. Sorting doesn't seem to give the right order. Still investigating why.

**Comment by mbarnett** [comment]
I think I have a fix for this. It was indexing GenericFiles in mixed case (instead of forcing to lowercase) but Collections in lowercase, and the sorting is case sensitive, so the results of a sort were inconsistent. 

The long-term fix would be to extend Solrizer to support alphaSort case-folding, but for the time being I'm just going to force GenericFile titles to lower-case for the sortable index.

**Comment by weiweishi** [comment]
Sounds good.  :+1:

File: app/controllers/concerns/hydranorth/collections/base_query.rb [hunk 4]
```diff
@@ -0,0 +1,25 @@
+module Hydranorth::Collections::BaseQuery
```
weiweishi: This is so much cleaner and easier to understand. :+1: @mbarnett

PR #1246, opened 2016-08-26
Fixes on Fixes on Fixes

mbarnett:
Fixes streaming, enables download links with DOIs indexed, enables title sorting in dropdowns, fixes default sorting on communities & collection.

Note that we need to reindex AGAIN to get filenames into the Solr index.

### DIFF
--- a/app/helpers/generic_file_helper.rb
+++ b/app/helpers/generic_file_helper.rb
@@ -20,10 +20,17 @@ module GenericFileHelper
     end
   end
 
-  def render_download_link(text = nil)
-    if @generic_file.label.present? || !@generic_file.filename.empty?
-      link_to (text || "Download"), download_path(@generic_file), { id: "file_download", target: "_new", data: { label: @generic_file.id } }
-    end
+  # if passed a block, this method conditionally renders the markup in the block only if the path is present. This
+  # provides an easy way of sanely dealing with some items having files (and thus download paths) and others not.
+  def render_download_link(item, text = nil)
+    # in Communities, Collections appear in lists alongside files, and so they are rendered through the same partial
+    # but collections have no download link
+    path = download_path(item)
+    return '' unless path.present?
+
+    download_link = link_to (text || 'Download'), path, { id: 'file_download', target: '_blank', rel: 'noopener noreferrer', data: { label: item.id } }
+    yield download_link if block_given?
+    return download_link
   end
 
   # sufia.download path is from Sufia::Engine.routes.url_helpers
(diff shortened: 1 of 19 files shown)


app/models/concerns/sufia/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
...
+      included do
+        contains "characterization", class_name: 'FitsDatastream'
+        property :mime_type,         delegate_to: 'characterization', multiple: false do |index|
+          index.as :stored_searchable
+        end
+        property :format_label,      delegate_to: 'characterization'
+        property :file_size,         delegate_to: 'characterization'
+        property :last_modified,     delegate_to: 'characterization'
+        property :filename,          delegate_to: 'characterization' do |index|
+          index.as :stored_searchable
```
mbarnett: N.B. we're overriding this file from Sufia so that we can index the filename to keep download link rendering fast with Peter's changes

mbarnett:
I'm about to push a couple of tweaks so maybe hold off on merging this for the next hour or so. I'll comment again when I'm done.

app/views/generic_files/_media_display.html.erb
```diff
@@ -0,0 +1,33 @@
+<% if @generic_file.image? %>
+
+  <%= render_download_icon "Download the full-sized image" %>
+
+<% elsif @generic_file.video? %>
+
+  <video controls="controls" class="video-js vjs-default-skin" data-setup="{}" preload="auto">
+    <source src="<%= download_path(@generic_file, file: 'webm') %>" type="video/webm" />
+    <source src="<%= download_path(@generic_file, file: 'mp4') %>" type="video/mp4" />
+    Your browser does not support the video tag.
```
mbarnett: We need to override this partial, because Sufia's version is hardcoded to call sufia.download_path, which breaks streaming with the new download routes

app/views/generic_files/_media_display.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  </video>
+
+<% elsif @generic_file.audio? %>
+
+  <audio controls="controls" class="audiojs" preload="auto">
+    <source src="<%= download_path(@generic_file, file: 'ogg') %>" type="audio/ogg" />
+    <source src="<%= download_path(@generic_file, file: 'mp3') %>" type="audio/mpeg" />
+    Your browser does not support the audio tag.
```
mbarnett: ditto

mbarnett:
This should be good to merge, now. I've made the filename behaviour more robust, so that it should run fine without a reindex, but when the reindex happens, it should run a little faster.

app/helpers/generic_file_helper.rb
```diff
@@ -33,41 +33,34 @@ def render_download_link text = nil
...
+    item = item_for_download(args.shift)
 
-    # not all doi_urls will be in SolrDocuments until a full reindex happens
-    if item.is_a?(SolrDocument) && !item.doi_url_indexed?
-      gf = GenericFile.find(item.id)
-      return gf.doi_url if gf.doi_url.present?
+    # doi supercedes anything else
+    return item.doi_url if item.doi_url.present?
+
+    # this shouldn't happen normally in production, because it indicates an item with no file
```
weiweishi: This actually does happen in production for 1. dataverse objects, 2. objects that are migrated from Thesis Deposit, where associating with a file is not mandatory.  So it may not necessarily be an error.

mbarnett: Good to know. I was basing that on Peter's original comment. Any idea what we should do for a download link in such cases? Right now it's just a circular link back to the file.

weiweishi: It doesn't display the link on the UI if there's no file attached. https://github.com/ualbertalib/HydraNorth/blob/master/app/helpers/generic_file_helper.rb#L23-L27. Maybe we should keep the logic the same? Not sure if item.filename.present? has the same effect as !item.filename.empty?

app/views/generic_files/_generic_file.html.erb
```diff
@@ -1,6 +1,7 @@
 <li>
   <h2>
     <%= link_to generic_file.title_or_label, sufia.generic_file_path(generic_file) %>
+    <small>[<%= link_to 'Download', download_path(generic_file), target: '_blank', rel: 'noopener noreferrer' %>]</small>
```
weiweishi: Should we have the same logic here - to not have the download link if there's no file attached?

app/helpers/generic_file_helper.rb
mbarnett: Ah, excellent. They do the same thing, but I'm going to push a change to have them perform the identical test anyways.

mbarnett: Actually, that will only work on item pages. For download links on search pages, that method isn't called, and won't work, since it assumes it's working with a GenericFile, and it needs to handle SolrDocuments on search and collection pages

app/views/generic_files/_generic_file.html.erb
mbarnett: Yeah, that's the conclusion I'm coming to. I'll need to refactor the helper to not assumer it's working with a generic file, since when this partial is called on search and collection pages, it's working with solr docs

weiweishi: Ay-ya-ya! Good catch. :+1:

mbarnett: Ok, I've found a couple of other dropdown menus that will need this change as well. I'll fix those and let you know when I've pushed a new version

mbarnett:
Ok, tests are passing, and I think this actually made some of the logic even cleaner. It should be good for another review now.

app/helpers/sufia_helper.rb
```diff
@@ -5,25 +5,21 @@ module SufiaHelper
 
   # override SufiaHelperBehavior method for institutional_access? condition
   def sufia_thumbnail_tag(document, options)
-    # collection
     if document.collection?
-      content_tag(:span, "", class: "glyphicon glyphicon-th collection-icon-search")
```
weiweishi: Out of curiosity, are double quotes only used in case of string interpolation in Ruby, and single quotes are always preferred?

weiweishi:
Looks good :+1:

app/helpers/sufia_helper.rb
mbarnett: It depends on who you ask. Rubocop tries to enforce single-quotes-unless-using-interpolation, and I guess I've internalized that because it makes us consistent with the rules the Hydra community would be following.

But it looks like [other people hate that](https://github.com/bbatsov/ruby-style-guide/issues/96) and [Github's styleguide disagrees entirely](https://github.com/styleguide/ruby/strings), so I guess the great thing about standards is there's so many to choose from.

PR #1247, opened 2016-08-29
Clean up uses of target= to ensure proper use of rel

mbarnett:
A quick PR to ensure we're using rel="noopener noreferrer" everywhere we're using target="_blank" on links. There are security risks whenever you use target="_blank" without the appropriate rel tags, and while not all of our link targets were external, it's easier to do this across the board, with no downside.

This blog post has a good explanation of the security issues: https://dev.to/ben/the-targetblank-vulnerability-by-example

(This PR also standardizes all of the links on target="_blank". Many were using target="_new", but [that's not valid HTML](http://stackoverflow.com/questions/4964130/target-blank-vs-target-new))


weiweishi:
👍 Thanks for the links. great info!

PR #1258, opened 2016-09-09
date_modified used the correct datetime value/zone, fixes #1061

### LINKED ISSUE #1061: date_modified used the wrong datetime value/zone when record is modified
@pbinkley has noticed that when a record is updated, the date_modified date is set incorrect, with current MST but set in UTC timezone. 

eg: an object updated on 11:01 MST shows date_modified as: Wed, 20 Apr 2016 11:01:30 +0000

Further investigation discovered that this is a bug in older version of Sufia:
https://github.com/projecthydra/sufia/blob/v6.2.0/app/controllers/concerns/sufia/batch_edits_controller_behavior.rb#L51

It uses Time.now instead of Time.current or Time.zone.now to get the time. 

This will affect OAI harvesting.

weiweishi:


app/controllers/concerns/hydranorth/batch_edits_controller_behavior.rb
```diff
@@ -11,6 +11,11 @@ def redirect_to_return_controller
         redirect_to sufia.dashboard_files_path
       end
     end
+    #This override is for Sufia 6.2. Issue is fixed in Sufia 7 and this will not be needed.
+    def update_document(obj)
+      super
+      obj.date_modified = Time.current.ctime
```
mbarnett: If anybody's curious, this is identical to the corresponding fix in Sufia 7: https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/batch_edits_controller_behavior.rb#L31

Issue #889, opened 2015-12-09
Aggregate statistics in dashboard not updating


sfbetz:
I don't think the stats in the dashboard are updating. According to this, admin has 0 views and 0 downloads - which is definitely not the case.
![screen shot 2015-12-09 at 12 30 54 pm](https://cloud.githubusercontent.com/assets/12801010/11696326/eac38bac-9e70-11e5-91c9-0b721206ac06.png)

pbinkley:
Decision: remove the aggregated stats from the display ("which have been viewed / downloaded"). We'll restore this in Sufia 7.

PR #1259, opened 2016-09-09
remove the aggregated stats from the dashboard display, temporarily a…

weiweishi:
…ddressed #889


app/views/dashboard/_index_partials/_stats.html.erb
```diff
@@ -4,12 +4,16 @@
...
     <tr>
-      <td>You have deposited <span><%= number_of_files %> files</span> which have been:
+      <td>You have deposited <span><%= number_of_files %> files</span> 
+<!-- This is temporarily removed because of issue #889. And it will be fixed and restored in Sufia 7 -->
+<!--
+which have been:
 
         <ul class="views-downloads-dashboard">
           <li>- viewed <span><%= @user.total_file_views %></span> times</li>
           <li>- downloaded <span><%= @user.total_file_downloads %></span> times</li>
```
mbarnett: Can we delete this? Or comment it out via Ruby ```<%# ... %> ``` tags? That way these call don't get run and the results rendered in the comments

weiweishi: Done.

PR #1268, opened 2016-09-16
attach collection facet restrictions to facet links on collection pages

mbarnett:


mbarnett:
1. Weird, that wasn't broken for me. Let me take another look at it.

2. I brought this up at a standup last week sometime but AFAICT there's no way to present facet results on the collection page without massive changes to Blacklight. It does so much in class initializers that when you try derive a second controller from it at the same time as Catalog controller the whole mess explodes -- tons and tons of missing callbacks on CollectionController, because they were inserted by Blacklight dynamically when the class initialized and that happens when CatalogController initializes (which due to the quirk of coming earlier lexicographically, happens first. If CollectionController initialized first, Catalog would start blowing up. Unless there's something obvious I'm missing?

pgwillia:
* I can't get the catalog test to fail just running it by itself.  Might be just me that one time.
* I had a feeling that the user experience concern was a nit pick but probably still worth explaining to stakeholders. I'm pretty sure this was in the category of do it if it's easy to fix, or else it's low priority.

mbarnett:
I was really hoping you knew something I didn't because it would be SO much easier *sad trombone*. Double-checking the tests on my end now.

pgwillia:
Finished in 36 minutes 34 seconds (files took 9.93 seconds to load)
427 examples, 0 failures, 7 pending

PR #1275, opened 2016-09-28
show user activity to that user or admin only resolves #179

### LINKED ISSUE #179: User page when not logged in
Can navigate to it, but throws error - should redirect to login page

pgwillia:


app/views/users/_profile_tabs.html.erb
```diff
@@ -0,0 +1,13 @@
+      <ul class="nav nav-tabs" id="myTab">
+        <li class="active"><a href="#contributions"><i class="glyphicon glyphicon-star"></i> Contributions</a></li>
+        <li><a href="#profile"><i class="glyphicon glyphicon-info-sign"></i> Profile</a></li>
+        <% if current_user && (current_user.admin? || @user == current_user) %>
```
mbarnett: Thinking out loud, but could we approach this with cancan instead?

something like ```if can? :view_activity```, and then add that logic to the ability file?

pgwillia: Good idea! It took a bit (@current_user @user were both empty User class objects?!) to figure out the ability logic that would work.  Here it is for review!

mbarnett:
Looks good but she would be merging this to a sprint branch instead?

pgwillia:
#179 was marked as bug.  My understanding is that all bugs go to master?  I can change it though.

mbarnett:
I guess I was thinking that only hotfixes that were getting their own deploy would go to master, and everything going out in a planned tag would be in that branch? It doesn't really matter though.

pgwillia:
Going to request merge to 1.7 tag.

PR #1291, opened 2016-10-11
Tag 1.6

pgwillia:
With changes discussed should be ready to tag and deploy 1.6
* ansible playbook add ```Rails.application.routes.default_url_options[:host] = '{{ ansible_fqdn }}'``` to config/environments/*
* reindex for sorting/faceting


pgwillia:
PDF's aren't getting their thumbnails -- I noticed this during my CentOS 7 experiment too. Still investigating

pgwillia:
same/same with master branch on plano.  era-test is still creating/displaying pdf thumbnails.  Same versions of convert (imagemagick and ghostscript) on all machines.  Likely something in master branch?!

pgwillia:
thumbnail exists and is accessible from Fedora.  The ?file=thumbnail seems to direct to the full file for the objects I've looked at.

PR #1303, opened 2016-10-20
github parameter uses the git protocol...

murny:
... which transmits data without encryption. Use https instead.

From Bundler: 

> http:// and git:// URLs should be avoided if at all possible. These protocols are unauthenticated, so a man-in-the-middle attacker can deliver malicious code and compromise your system. HTTPS and SSH are strongly preferred.

More Info can be found here regarding this:
http://bundler.io/v1.13/man/gemfile.5.html#GIT
http://bundler.io/v1.13/man/gemfile.5.html#GITHUB

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'google-api-client', '~> 0.7.1'
 
 gem 'sufia', '~> 6.2.0'
 gem 'rsolr', '~> 1.0.6' # blacklight will not load by default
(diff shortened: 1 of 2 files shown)


pgwillia:
What's your take on including/excluding Gemfile.lock from repos? We've found it useful to include so we can see exactly which version of gems were used when debugging the 'well it works on my machine' situations.

murny:
Haha yes, I perfer to include the lock file. Just forgot it, oops! Good call! 

Thanks!

Issue #1301, opened 2016-10-17
DOI should appear in endnote export


pgwillia:
#1276

display on endnote export
- public [DOI](https://github.com/ualbertalib/HydraNorth/commit/091ec57ea85351b28197b2a9d38af21344d224c2) `Rails.application.routes.url_helpers.persistent_id_url(id)`
- otherwise [current](https://github.com/ualbertalib/HydraNorth/blob/31a0e74aeb3a877b1786d3473ffe84a7d21b01c3/app/models/concerns/hydranorth/generic_file/export.rb#L6) `Rails.application.routes.url_helpers.generic_file_url(id)`

Issue #1284, opened 2016-09-28
DOI label on the object display page


weiweishi:
- [ ] To display the DOI label on object page, and other places as stakeholders specified.

sfbetz:
![2016-09-29_16-22-41](https://cloud.githubusercontent.com/assets/12801010/18974663/ae41d092-8661-11e6-87f8-d73c374bdf72.png)

sfbetz:
Requirements:
-remove "Actions" Heading
-Move "Analytics" link up below "download the full-sized PDF"
-Add Share heading (or rename "Actions")
-add Permanent link (DOI): above other share options. Populate field with DOI link 

-Add field below link for ERA DOI: populate field with DOI.

We don't need DOI in the search results list.

pgwillia:
#1276 
Permenant link may be nil for some items (i.e. Dataverse), so this section should be hidden. DOI may also be 'unavailable' for some items (i.e. Private items) and can still be displayed to the user and admin.

sfbetz:
The DOIs should be displayed like this: https://doi.org/10.1629/22161

CrossRef (a major DOI registration agency) has updated their documentation to drop the .dx from the DOI url. This will become the new standard for published article DOIs, so one we should probably follow as well.

murny:
Example of what this label change will look like (if there was a PDF link it would be above the analytics):
![image](https://cloud.githubusercontent.com/assets/1930474/20077087/29447b28-a4f8-11e6-948e-fcdb92730438.png)

Regarding:
-Add field below link for ERA DOI: populate field with DOI.

Assuming there is nothing to do for this anymore?

leahvanderjagt:
Looks good.

___________________________________________________________




*Leah VanderjagtDigital Repository Services Coordinator University of
Albertat. 780.492.3851 *
*leahv@ualberta.ca <leahv@ualberta.ca> *

murny:
After talking to Leah, I have added a field for DOI under the descriptions section in between "Date created" and "License information". Example of what this might look like:
![image](https://cloud.githubusercontent.com/assets/1930474/20079571/8c87b8bc-a502-11e6-9b2d-4bf49f08ce87.png)

Issue #1283, opened 2016-09-28
DOI Unavailable


weiweishi:
- [ ] Based on output of #1280 
- [ ] Handles the state change of DOI. (unavailable)
- [ ] Set [status message](http://ezid.cdlib.org/doc/apidoc.html#identifier-status) (reason for the object's unavailability) in EZID (To provide information on the registry)

pgwillia:
#1276 
There are two cases for unavailable: deleted items and private items.
Both should set ```_export=no```

johnhuck:
From @leahvanderjagt and @sfbetz, the status message for private items should be "not publicly released"

johnhuck:
I can now confirm that we recommend using the :erc code `:unal` , which means "unallowed; intentionally suppressed" in place of a value for each of the required metadata elements when registering private items as with a DOI status of "unavailable."

When an item is marked "unavailable" because it is deleted, however, the metadata in its last public state should be maintained. That is, if the item were public, it should retain normal metadata, as a guide to people who may have the DOI. If the item were private before it was deleted, though, the `:unal` codes should be retained, since we don't want to advertise its identity after the fact.

Please note that this code is different from the :erc code `:unav` ("unavailable; possibly unknown") which we have recommended using for normal items that lack a four digit year of publication.

The status message to accompany the chance in status to unavailable for private items should be "unavailable | not publicly released".

For deleted items, I am going to recommend "unavailable | withdrawn" .

Let me know if there are any other questions I can help with. Cheers!

Issue #1282, opened 2016-09-28
Modifying DOIs


weiweishi:
- [ ] Modify DOI metadata if metadata is changed on the object
- [ ] Based on the output of #1280

PR #1309, opened 2016-11-04
#1264 DOI Implementation

murny:
This PR covers the following Issues for milestone 1.7:

#1281
#1282
#1283
#1284
#1301

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,12 @@ source 'https://rubygems.org'
 # Use database to store sessions
 gem 'activerecord-session_store'
 
+gem 'aasm'
+gem 'aasm-active_fedora'
+gem 'ezid-client'
+
+gem 'resque'
+
 # Avoid cannot load such file -- google/api_client
 gem 'google-api-client', '~> 0.7.1'
 
@@ -75,6 +81,8 @@ gem 'rest-client'
 gem 'noid', '~> 0.8'
 
 group :test do
+  gem 'vcr', require: false
+  gem 'webmock', require: false
   gem "rspec-its"
   gem "ci_reporter_rspec"
   gem "timecop"
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Relative to the application directory
 
 To Restart Components
 ---
-The shell script `bin/resta. You will still be required tort-all` runs these commands:
+The shell script `bin/restart-all` runs these commands:
 * Jetty
  * ```cd /var/www/sites/hydranorth && rake jetty:restart```
 * Resque/Redis
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Relative to the application directory
 
 To Restart Components
 ---
-The shell script `bin/restart-all` runs these commands:
+The shell script `bin/resta. You will still be required tort-all` runs these commands:
 * Jetty
  * ```cd /var/www/sites/hydranorth && rake jetty:restart```
 * Resque/Redis
@@ -91,9 +91,9 @@ The shell script `bin/reset-all` runs these commands:
 Batch ingest
 ---
 
-- **migration:user_migration** which migrates users in the user files (data derived from ERA mysql tables user and author) use: ```rake migration:user_migration['lib/tasks/migration/test-metadata/users.txt']``` 
-  - **note: this has only been tested with small dataset - need to proceed with caution due to the potential for sending emails to migrated users** 
-  - TO-DO: migrate user avator/profile images 
+- **migration:user_migration** which migrates users in the user files (data derived from ERA mysql tables user and author) use: ```rake migration:user_migration['lib/tasks/migration/test-metadata/users.txt']```
+  - **note: this has only been tested with small dataset - need to proceed with caution due to the potential for sending emails to migrated users**
+  - TO-DO: migrate user avator/profile images
 - **migration:era_collection_community** which migrates collections and communities in the metadata directory use (run community migration first then collection migration): ```rake migration:era_collection_community['lib/tasks/migration/test-metadata/community']```
 ```rake migration:era_collection_community['lib/tasks/migration/test-metadata/collection']```
 - **migration:eraitem** which migrates active/non-deleted items from the metadata directory (argument from the rake task) use: ```rake migration:eraitem['lib/tasks/migration/test-metadata']```
@@ -101,7 +101,7 @@ Batch ingest
   - **note: migration has to happen in the following order: communities, collections, then eraitems.**
   - **note: file name should start with "uuid_", only those files will be selected.**
 - ```rake hydranorth:update_special_itemtype``` will update the resource type "report" to "computing science technical report" if this item is a member of "technical report". In order for the rake task to work, the collection has to be migrated already and exist in the system.
-- ```rake hydranorth:characterize``` will push all the items to the characterize resque pool for characterization, and thumbnail creation. This should be done after a complete fresh migration - as currently the migration job disables the resque jobs for faster completion. 
+- ```rake hydranorth:characterize``` will push all the items to the characterize resque pool for characterization, and thumbnail creation. This should be done after a complete fresh migration - as currently the migration job disables the resque jobs for faster completion.
   - **note: ```rake hydranorth:characterize_some['filename']``` will push the items in the given list to the characterize resque job. **
 - **batch:ingest_csv** is used by ERA Admin and ERA Assistants to batch ingest from a csv file. Takes the csv file and a directory where the referenced files exist. use: ```rake batch:ingest_csv['lib/tasks/batch/ERA_batch_ingest.csv','lib/tasks/batch/files_and_metadata/']```
   - **note: collections and communities dependencies must exist.**
@@ -109,9 +109,9 @@ Batch ingest
 Ingest Dataverse Metadata
 ---
 
-- **migration:clean_up_withdrawn[old_dir,new_dir]**: compare the last ingest directory and the current ingest directory, detect and clean any study that has been withdrawn from Dataverse. 
+- **migration:clean_up_withdrawn[old_dir,new_dir]**: compare the last ingest directory and the current ingest directory, detect and clean any study that has been withdrawn from Dataverse.
   - **note: The Dataverse metadata export only includes current published studies, and doesn't include any information about withdrawn studies. Ingest directories are always named with timestamps.**
-- **migration:dataverse_objects[dir]**: ingest the metadata for dc term files in the given directory. It will create new objects for new studies, and update metadata for existing studies. 
+- **migration:dataverse_objects[dir]**: ingest the metadata for dc term files in the given directory. It will create new objects for new studies, and update metadata for existing studies.
   - use: ```rake migration:dataverse_objects['spec/fixtures/migration/test-metadata/dataverse/']```
 
 
@@ -120,13 +120,13 @@ Generate Sitemap
 ** for Google Scholar **
 - use: ```rake sitemap:generate```
 
-Populating Local Geonames Authority 
+Populating Local Geonames Authority
 ---
-- **hydranorth:harvest:geonames_cities** which downloads the latest list of cities1000 file from geonames.org, and populates the data into local database tables: local_authorities, and local_authority_entries. 
+- **hydranorth:harvest:geonames_cities** which downloads the latest list of cities1000 file from geonames.org, and populates the data into local database tables: local_authorities, and local_authority_entries.
   - use: ```rake hydranorth:harvest:geonames_cities```
   - file downloaded: http://download.geonames.org/export/dump/cities1000.zip
   - scope for authority entries: contains all cities with a population >1000 or seats of adm div (ca 80.000)
-  - local_authority_entries includes labels and URIs. Currently labels are used for autocompletion and saved in the record. 
+  - local_authority_entries includes labels and URIs. Currently labels are used for autocompletion and saved in the record.
 
 Configuring Shibboleth
 ---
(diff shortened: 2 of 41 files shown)


app/actors/hydranorth/generic_file/actor.rb
```diff
@@ -18,6 +18,8 @@ def update_metadata(attributes, visibility)
         date_created = attributes[:date_created]
         generic_file.year_created  = date_created[/(\d\d\d\d)/,0] unless date_created.nil? || date_created.blank?
 
+        # TODO Do this much smarter! If you just hit an update from the UI,
+        # all this logic runs even though nothing changes! Should be checking for a diff of the ID's?
         ['hasCollectionId', 'belongsToCommunity'].each do |attr|
```
murny: I noticed this code block was executing on every update of a GenericFile from the UI. This is because when a user hits the update button, it will serialise the entire form and all its attributes and send it to our controller. We simply make the assumption here that if the attributes contains hasCollectionID or belongsToCommunity then they must have changed. But this simply isn't the true.

This does a lot of IO to the database (a few lookups and a few saves), which is expensive with Fedora. We should be doing a simple check to see if they actually changed before proceeding any further.

app/jobs/doi_removal_job.rb
```diff
@@ -0,0 +1,7 @@
+class DOIRemovalJob < ActiveJob::Base
```
murny: I pulled in resque and using activejob API (comes with rails 4.1+), instead of hooking into Sufia's job queue like we have done with app/jobs/batch_update_job.rb. Why?

Sufia's job approach is a custom api on top of resque which seems to be entirely replaced in Sufia 7+. Why not follow this approach? Well, for one they are marshalling the entire job in the queue which loses visibility and information about the job. As a result they are naming their jobs as queues, which isn't the purpose of a queue. For example:
```
Sufia::Resque::MarshaledJob    ["BAhjEkRPSU1pbnRpbmdKb2I=\n"]
```
vs
```
ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper    [{"job_class"=>"DOIRemovalJob", "job_id"=>"e5f88c8a-3f28-435e-902d-a340ee36194d", "queue_name"=>"default", "arguments"=>["doi:10.5072/FK2ZW1FW18"], "locale"=>"en"}]
```
Also they are not following the standard resque API (for example using a run method instead of perform method which seems to have no documentation about these changes, etc.). With Sufia 7.0+, they have removed this approach entirely and simply uses ActiveJob API which is great! So for sake of future proofing this and doing this better, I decided to use ActiveJob API instead.

app/models/concerns/hydranorth/generic_file/doi_states.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        def doi_permanent_url
+          "https://doi.org/#{doi.gsub(/doi:/, '')}" if doi.present?
+        end
+
+        def doi_fields_present?
+          # TODO: Shouldn't have to do this as these are required fields on the UI.
+          # However since no data integrity a GF without these fields is technically valid... have to double check
+          self && title.present? && creator.present? && resource_type.present? && Sufia.config.admin_resource_types[resource_type.first].present?
```
murny: As comment said, this shouldn't be necessary. But seems like we have no integrity around the data for a GF. Technically the only required field from the back end is a depositor must have edit access (depositor/edit_users fields cant be nil)/ However, from the UI , the required fields are the following: title, creator, resource type, language, subject/keyword, license, and community.  But the only thing validating that these required fields are being set is an html5 required attribute on the fields (and not all browsers support html5 required attributes)! So because of this, I am double checking that the fields required for the DOI service are actually present on the specific GF.

config/environments/production.rb
```diff
@@ -85,3 +85,6 @@
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 end
+
+# Required when using Rails.application.routes.url_helpers from outside the request/response life cycle (models, jobs, lib)
+Rails.application.routes.default_url_options = { host: 'era.library.ualberta.ca', port: 443, protocol: 'https' }
```
murny: We are rewriting/generating these files from ansible if my understanding is correct. Is there a reason we are doing this instead of using environmental variables?

config/secrets.yml
```diff
@@ -12,11 +12,18 @@
...
+  ezid_user: apitest
+  ezid_password: apitest
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
 production:
-  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
+  secret_key_base: <%= ENV['SECRET_KEY_BASE'] %>
+  ezid_default_shoulder: <%= ENV['EZID_DEFAULT_SHOULDER'] %>
+  ezid_user: <%= ENV['EZID_USER'] %>
+  ezid_password: <%= ENV['EZID_PASSWORD'] %>
```
murny: Rails 4.1 introduced a very easy way to manage environmental variables in your app instead of using an external gem to do this. Simply define your secret keys in your environments, then simple call 

```
Rails.application.secrets.<key>
```

In your code.

spec/lib/hydranorth/doi_service_spec.rb
```diff
@@ -0,0 +1,100 @@
+require 'spec_helper'
+
+describe Hydranorth::DOIService do
+  let(:example_doi_id) { 'doi:10.5072/FK2JQ1003W' }
+  let(:generic_file) do
+    # TODO this is coming from sufia's factories, fix this
```
murny: I feel like we should have a standard generic file factory. We have before(each)/let blocks through out our test suite doing very similar logic for generating new Generic Files. Should be consolidated. Also we are including all the factories from Sufia as well. Not sure why? Here I am hooking into Sufia's generic file factory, but this should be using our own.

spec/models/generic_file_spec.rb
```diff
@@ -75,6 +70,7 @@
   describe '#append_metadata', :integration => true do
 
     before  do
+      # TODO: Don't use a before each, use a let instead
```
murny: This appears to be a common issue in our test suite, is using a before each block to define objects for our test cases. Should be using a let block instead as it's faster performance wise. Let blocks are lazily loaded and only used when referenced.

spec/spec_helper.rb
```diff
@@ -12,6 +12,14 @@
 require 'factory_girl_rails'
 require File.join(Sufia::Engine.root, 'spec/factories/generic_files')
 require File.join(Sufia::Engine.root, 'spec/support/features')
+
```
murny: With Rspec v3, we shouldn't be using the spec_helper for our main test helper. spec_helper.rb is suppose to be a lighter weight test helper for use in a non rails environment (to save on boot time). The rails_helper.rb is suppose to have everything needed for our rails tests.

Also, instead of polluting this file with everything we need to have configured for our tests, these configurations and helpers should live in their own files under support directory. This seems to be a best practice for better code readability. You could even take it a step further and remove this globbing of files defined below and only include the files for certain tests (for example capybara should only be initialised on integration tests for example). Something to think about.

config/environments/production.rb
murny: I may be biased as a developer but from your list of pros and cons, the pros heavily out weigh the cons. 

The con is a valid concern, but I feel like this is where good naming convention/documentation of env variables will help. And the only things that should be moved to ENV variables are things you don't want to be made public to the world (API keys, Passwords, Usernames, etc etc) or maybe some dynamic configuration depending on servers or something.

Also it looks like we are already using env variables in these config files in the ansible playbook anyways? So not sure if this is a con if this is how we are currently doing it in ansible. 

The main concern I guess is why do we require the generation/replacing of files in the ansible playbook when you can just simply use the ones in the rails app repo? It gets confusing when you have a file in your rails repo that you think is going to production but it's actually doing nothing as it is getting replaced by a file from ansible (which now someone has to update and maintain this code in ansible). Like the production.rb in ansible (assuming i'm looking in the right place) only has three ENV VARs (DEFAULT_URL_OPTIONS, APP_EMAIL, CONTACT_EMAIL). The rest of the file isn't sensitive. You could just update the production.rb in this repo and just copy it over and that's it? No need to have another copy living in ansible that is generated and then copied over and replacing the one in this repo when a deployment happens?

That's just my 2 cents. ¯\ _ (ツ) _ /¯

pgwillia: The production.rb in ansible is using variables, not ENV VARs for replacement.  A minor point of clarification for sure. I agree that a good direction would be to set the values in the environments configs to be environment variables the way you described for the secrets and then have the ansible playbook set their values. Makes sense to me!

Gemfile
```diff
@@ -67,14 +73,16 @@ gem "clamav"
...
-gem 'sitemap', github: 'ualbertalib/rails-sitemap'
+gem 'sitemap', git: 'https://github.com/ualbertalib/rails-sitemap.git'
 
 # to fetch noid from fedora for reindex job
 gem 'rest-client'
 
 gem 'noid', '~> 0.8'
 
 group :test do
+  gem 'vcr', require: false
```
pgwillia: I'd not heard about vcr before. Is there anything to keep back of mind when using it?

murny: It's for mocking external API's for testing. In our case EZID. 

So when running your test suite, VCR will intercept any external HTTP request and do the following:

- On the first test run, it will do a real HTTP request to the external service/API (EZID), and record the response/request and cache it in your repo (which can then be version controlled and checked into github). 
- When you run it again, VCR will replay the response from the cached copy instead of doing a real HTTP request again. This makes your test fast (no real HTTP requests are made anymore), deterministic (the test will continue to pass, even if you are offline, or external API goes down) and accurate (the response will contain the same headers and body you get from a real request).  

More info:

https://github.com/vcr/vcr
https://relishapp.com/vcr/vcr/docs

pgwillia:
```
Failures:

  1) Hydranorth::Forms::GenericFileEditForm#terms should return a list
     Failure/Error: expect(subject.terms).to eq([:title, :creator, :contributor, :subject, :resource_type, :language, :spatial, :temporal, :description, :date_created, :license, :rights, :is_version_of, :source, :related_url, :belongsToCommunity, :hasCollectionId ])
       
       expected: [:title, :creator, :contributor, :subject, :resource_type, :language, :spatial, :temporal, :description, :date_created, :license, :rights, :is_version_of, :source, :related_url, :belongsToCommunity, :hasCollectionId]
            got: [:title, :creator, :contributor, :subject, :resource_type, :language, :spatial, :temporal, :description, :date_created, :doi, :license, :rights, :is_version_of, :source, :related_url, :belongsToCommunity, :hasCollectionId]
       
       (compared using ==)
       
       Diff:
       @@ -8,6 +8,7 @@
         :temporal,
         :description,
         :date_created,
       + :doi,
         :license,
         :rights,
         :is_version_of,
     # ./spec/forms/generic_file_edit_form_spec.rb:8:in `block (3 levels) in <top (required)>'

Deprecation Warnings:

Using `should` from rspec-expectations' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` with `config.expect_with(:rspec) { |c| c.syntax = :should }` instead. Called from /var/www/sites/hydranorth/spec/tasks/dataverse_migration_rake_spec.rb:89:in `block (3 levels) in <top (required)>'.


If you need more of the backtrace for any of these deprecations to
identify where to make the necessary changes, you can configure
`config.raise_errors_for_deprecations!`, and it will turn the
deprecation warnings into errors, giving you the full backtrace.

1 deprecation warning total

Finished in 69 minutes 49 seconds (files took 16.39 seconds to load)
446 examples, 1 failure, 5 pending
```

murny:
Good catch on the failing test @pgwillia! I'll get that fixed shortly. I did do a full test run before I committed this, not sure why it was still succeeding on my local machine :confused:

config/secrets.yml
```diff
@@ -12,11 +12,18 @@
 
 development:
   secret_key_base: 5091e17fd4c0a3ddacdc134a3926e29c464cdaf0526684912fcd61f89c30680ec5be714f3f66b2c6c7ed6c5384b60eaf31584e10b9cff8e7bdaf9e2d8d6d11ab
-
+  ezid_default_shoulder: doi:10.5072/FK2
+  ezid_user: apitest
+  ezid_password: apitest
```
pgwillia: In the ezid-client documentation they instruct that the test user password is not provided and EZID should be contacted. We probably shouldn't be publishing.

app/views/generic_files/_descriptions.html.erb
```diff
@@ -3,7 +3,7 @@
         <%= hidden_field_tag('redirect_tab', 'descriptions') %>
         <h1>Descriptions <span class="pull-right required"><abbr title="required">*</abbr> indicates required fields</span></h1>
         <div class="well">
-            <% form_terms = f.object.terms - [:degree_grantor, :date_submitted, :date_accepted] %>
+            <% form_terms = f.object.terms - [:degree_grantor, :doi, :date_submitted, :date_accepted] %>
```
murny: Shouldn't have variable assignment logic in the view, should be done in the GenericFileEditForm class.

This is required so DOI isn't an editable field in the descriptions.

config/initializers/inflections.rb
```diff
@@ -14,3 +14,8 @@
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.acronym 'RESTful'
 # end
+
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.acronym 'DOI'
```
murny: Required when `.humanize` is called on doi, it will become `DOI` instead of `Doi`

app/views/generic_files/_descriptions.html.erb
mbarnett: Well if you're going to touch it, you might as well clean it up 😉

README.md
```diff
@@ -54,7 +54,7 @@ Relative to the application directory
 
 To Restart Components
 ---
-The shell script `bin/restart-all` runs these commands:
+The shell script `bin/resta. You will still be required tort-all` runs these commands:
```
pgwillia: typo?

spec/forms/generic_file_edit_form_spec.rb
```diff
@@ -1,11 +1,9 @@
 require 'spec_helper'
 
 describe Hydranorth::Forms::GenericFileEditForm do
-  subject { described_class.new(GenericFile.new) }
-
   describe "#terms" do
```
murny: This is an extremely poor test. #terms is a method being inherited from the presenter class. The presenter test should be testing this logic (which we have no tests for), and GenericFiledEditForm specific tests should be in this file?

spec/forms/generic_file_edit_form_spec.rb
```diff
@@ -1,11 +1,9 @@
...
     it "should return a list" do
-      expect(subject.terms).to eq([:title, :creator, :contributor, :subject, :resource_type, :language, :spatial, :temporal, :description, :date_created, :license, :rights, :is_version_of, :source, :related_url, :belongsToCommunity, :hasCollectionId ])
+      expect(Hydranorth::Forms::GenericFileEditForm.terms).to eq([:title, :creator, :contributor, :subject, :resource_type, :language, :spatial, :temporal, :description, :date_created, :doi, :license, :rights, :is_version_of, :source, :related_url, :belongsToCommunity, :hasCollectionId ])
```
murny: Highly against subject syntax, especially when its confusing magic like this one.

```subject { described_class.new(GenericFile.new) }
subject.terms
isn't intuitive on what is going on.
```

mbarnett: 👍

mbarnett:
🎉 

What needs to happen for deployment? 
[ ] Get the credentials for the prod ezid account into ansible? 
[ ] Run rake tast to set status on existing items post-deploy?

Anything else?

PR #1311, opened 2016-11-09
Code cleanup

murny:
Low hanging fruit of some code cleanup I've noticed.

- Remove test directory, as we not using Test::Unit (MiniTest) but using Rspec instead
- Remove .keep files that are not needed and README.rdoc (as we are using README.md instead)
- Move external js libraries to proper place - See http://guides.rubyonrails.org/asset_pipeline.html#asset-organization for more info
- Tiny cleanups on Gemfile (Removal of a few default gems not being used)

Some interesting profiling statistics of our boot up performance of our current app:

Loading of the rails environment benchmarks (On a VM):

```
0.101114305 rails/railtie
0.10182317 action_view/helpers/form_tag_helper
0.104302322 rails/engine
0.1046031 active_record
0.105816881 parslet/atoms
0.106169747 sufia/models
0.106212778 byebug/core
0.106300596 google/api_client/service_account
0.109562689 sass/engine
0.110274024 active_record/session_store/session
0.112500639 cancan/controller_additions
0.119835944 legato
0.121055356 cancan
0.123099815 sprockets
0.127088541 sass
0.128917932 bootstrap-sass
0.132420291 active_record/session_store
0.137814365 sprockets/rails/context
0.139030662 blacklight/engine
0.147672505 rails/application
0.149233231 active_record/railtie
0.153016911 mini_magick
0.153696725 blacklight
0.168654337 oauth/client/helper
0.207452942 oauth
0.208520915 rails
0.223387825 rails/tasks
0.248365865 parslet
0.248491763 rdf/vocab
0.252184334 mail/field
0.254848374 parsing_nesting/grammar
0.263361541 parsing_nesting/tree
0.2690505 sprockets/railtie
0.270367664 blacklight_advanced_search/parsing_nesting_parser
0.33358607 sufia/models/active_fedora/redis
0.366799898 google/api_client
0.401136876 linkeddata
0.428882168 ldp
0.449430931 active_fedora
0.449948289 active-fedora
0.497377611 mail
0.572525719 sufia/analytics
0.577648122 /var/www/sites/hydranorth/config/environments/development.rb
0.646229 rails/all
0.728615688 hydra-access-controls
0.729478395 hydra-core
0.729788601 hydra/head
11.780763837 /var/www/sites/hydranorth/config/environment.rb

real	0m15.955s
user	0m6.030s
sys	0m8.029s
```
Other interesting Information:
```
Slow requires (gems):
   103.98  byebug
   114.61  pdf-reader
   146.03  bootstrap-sass
   163.03  pry
   227.89  rails
  1087.91  sufia
```
```
Slow requires (initializers):
    172.81  :build_middleware_stack
    243.90  blacklight.helpers
    328.04  patches
    592.04  :set_clear_dependencies_hook
    632.79  :load_environment_config
    667.09  requires
    752.76  :set_routes_reloader_hook
    973.54  :run_prepare_callbacks
    997.35  :finisher_hook
   5742.89  ./config/initializers/clamav.rb
   5879.27  :load_config_initializers
```
 This pull request will remove two of the slow gem requires that pop out which are pdf-reader and bootstrap-sass. PDF-Reader is only used in a rake task (so by using a `require: false` in the gemfile it will not be loaded on boot and only will be loaded when required (when someone runs the rake task). Bootstrap-sass is never used in our app. So I can safely remove this.

What does all this tell us? Seems to be our two big bottlenecks is sufia and the intiailization of ClamAV. We can't really do to much about sufia. However I feel like we can definately do something about ClamAV. Is ClamAV needed in development and test environments for any reason? (at the very least it should be scoped to linux environments as it depends on an external library which seems to be a pain to install on mac/windoze) . By simply removing the ClamAV initializer in the test enviornment, we see a major speed up of  5-6 seconds for boot time on my VM!

Examples of running a single spec (`$ bundle exec rspec spec/lib/hydranorth/writable_permissions_spec.rb` )

Before:
```
Finished in 0.25495 seconds (files took 16.12 seconds to load)
```
After:
```
Finished in 0.22909 seconds (files took 9.8 seconds to load)
```

Other questions:

- `spec-views` directory needed? Could probably be tossed
- Gems that potential could be removed:
  - `google-api-client` - Can't see this being used in our code base (underlining gems have it as a dependency but we don't. I tried removing it and everything seemed to work fine?)
  - `coffee-rails` - We are not using coffeescript anywhere in our app. Shouldn't need this. (I seem to get a load error when removing it, might be caching issue on my VM however)
  -  `devise-guests` - We do not allow guest logins and I see no usage of the helper methods provided by this gem in our code base
  - `activerecord-session_store` - Kinda disagree with the usage of this gem. Looks like this was brought in for a potential security issue with cookie session based storage which shouldn't be a concern if your using sessions properly. As of Rails 4, cookies are signed and encrypted by the app server and it's contents cannot be viewed or tampered with by the client. The biggest security concern with cookie session storage is what is referred to as a "Reply attack". Which only is a risk if you are improperly storing sensitive information in the users session (which shouldn't happen regardless, if for example you have sensitive information for a user, you should be storing the users_id in the cookie and the sensitive information in the users table instead which can be looked up by the user_id).  More information about what a Replay Attack is can be found here : http://guides.rubyonrails.org/security.html#replay-attacks-for-cookiestore-sessions. The problem with storing the session in the database is it puts an unnecessary load on your database. Once your application receives a large amount of traffic, the sessions database table is continuously bombarded with read/write operations. There is probably a good reason why as of Rails 4, the Active Record session store was removed from the core framework and is now deprecated. Also another concern about this is, are we ever cleaning up/maintaining this table and clearing out old sessions in production? I don't think we have much traffic but this table could potentially be enormous?


Gemfile
```diff
@@ -19,43 +19,16 @@ gem 'mysql2', '~> 0.3.2'
...
 gem 'coffee-rails', '~> 4.1.0'
 
-# See https://github.com/sstephenson/execjs#readme for more supported runtimes
-# gem 'therubyracer', platforms: :ruby
-
 # Use jquery as the JavaScript library
 gem 'jquery-rails'
 
-# Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
-gem 'turbolinks'
```
murny: We are not using turbolinks at all in our app - removed

Gemfile
```diff
@@ -19,43 +19,16 @@ gem 'mysql2', '~> 0.3.2'
...
-
-# Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
```
murny: Same with jbuilder and sdoc?

Gemfile
```diff
@@ -19,43 +19,16 @@ gem 'mysql2', '~> 0.3.2'
 
 # Use SCSS for stylesheets
 gem 'sass-rails', '~> 5.0'
-gem 'bootstrap-sass', '~> 3.3.4.1'
```
murny: Cannot see this being used anywhere. Seems like we have our own customized css (sufia might be using it but that's their own dependencies)

Gemfile
```diff
@@ -64,7 +37,7 @@ gem "omniauth-shibboleth"
 gem "clamav"
 
 # for migration reading the license file
-gem "pdf-reader"
+gem "pdf-reader", require: false
```
murny: Only needed in a rake task (which already has it required in that task file), shouldn't be loaded in rails environment

Gemfile
```diff
@@ -76,8 +49,9 @@ gem 'noid', '~> 0.8'
 
 group :test do
   gem "rspec-its"
-  gem "ci_reporter_rspec"
+  gem "ci_reporter_rspec", require: false
```
murny: ci_reporter_rspec and brakeman are CLI? shouldn't be initialized on load

Gemfile
```diff
@@ -91,12 +65,14 @@ group :development, :test do
...
-  gem "show_me_the_cookies"
-  gem "brakeman"
+  gem 'brakeman', require: false
   gem "pry"
   gem "pry-remote"
   gem 'pry-byebug'
+  gem 'ruby-prof'
+end
+
+group :development do
```
murny: Don't need better_errors in test environment

app/views/layouts/application.html.erb
```diff
@@ -2,8 +2,8 @@
 <html>
   <head>
     <title>Hydranorth</title>
-    <%= stylesheet_link_tag    "application", media: "all", "data-turbolinks-track" => true %>
```
murny: Don't think this application layout is ever used, but cleaning up the data-attrs for turbolinks.

Gemfile
```diff
@@ -7,6 +7,7 @@ gem 'activerecord-session_store'
 gem 'google-api-client', '~> 0.7.1'
 
 gem 'sufia', '~> 6.2.0'
+gem 'jbuilder', '~> 2.0' # hidden sufia dependency
```
murny: Sufia appears to be using Jbuilder but they do not have it as a dependency :disappointed:  ? Am I missing something here? Without this in our Gemfile, creating a GF will fail from the UI.

PR for adding usuage of jbuilder: https://github.com/projecthydra/sufia/pull/797
Jbuilder file in sufia 6.2 that is causing the issue: https://github.com/projecthydra/sufia/blob/c9cbfdbd7b52619aaf848a3d3b9a984c452a0481/app/views/generic_files/jq_upload.json.jbuilder

Even in current Sufia, this problem appears to still exist (see curation concerns/sufia).

Gemfile
pgwillia: cispec is a rake task but brakeman is CLI -- both used http://cardiff.library.ualberta.ca/job/HydraNorth

pgwillia: Is there a better way to do this? https://github.com/ualbertalib/HydraNorth/blob/master/lib/tasks/spec.rake

Gemfile
pgwillia: One thing to keep in mind is that we currently use the "test" environment for our era-test staging servers. I think this should change to use a "staging" environment but requires buy in and changes from ops. I agree that we don't need better_errors in test environment, but something to keep in mind.

murny: Then yeah you guys should have a staging environment for these UAT servers. As rails "test" environment is quite different. It's super easy to set up too. Have a staging.rb in your environments directory (should mimic production as much as possible, maybe with different email configuration and what not) and use a RAILS_ENV=staging instead. I will say though, you should be serving the proper 404.html/500.html error pages on staging like you would on production and not using better_errors?

Gemfile
mbarnett: They're terrible at listing their dependencies properly. I've hit similar issues multiple times.

Gemfile
murny: This is a valid point if we don't have access. But also I say that cause I notice on production we are using the default error pages (except for the 404 page) coming out of Rails instead of doing something custom? We don't even have a favicon or anything either in prod. Really doesn't matter to me honestly if we keep it in or not.

mbarnett: Yeah, we don't currently have custom designs for most error pages, and likely won't until @nnunn is back from mat leave.

murny: Custom design be icing on the cake (and they don't need to be fancy haha), but even maybe more useful information/language on them wouldn't be a terrible idea? Like `if the error condition persists, you can contact us at <SOME EMAIL ADDRESS> with additional detail, including steps to reproduce the error....blah blah`.

pgwillia:
@murny I can't remember.  Is this ready for review and merge?

murny:
Yep this is ready for review! All tests are passing in this pull request.

pgwillia:
Haven't figured out why yet, but seeing these failures:
```
Failures:

  1) collection modify collection and community should remove Official and Community flags
     Failure/Error: Unable to find matching line from backtrace
     ActionView::Template::Error:
       invalid slice size
     # ./app/views/communities/index.html.erb:2:in `each_slice'
     # ./app/views/communities/index.html.erb:2:in `_app_views_communities_index_html_erb__688731271463027035_33598480'

  2) generic file new new item fields should not allow multiple resource_type selections, but assign to an array
     Failure/Error: expect(page).to have_xpath('//select[@name="generic_file[resource_type][]" and not(@multiple)]')
       expected to find xpath "//select[@name=\"generic_file[resource_type][]\" and not(@multiple)]" but there were no matches
     # ./spec/forms/generic_file_new_form_spec.rb:55:in `block (3 levels) in <top (required)>'

Finished in 29 minutes 57 seconds (files took 3 minutes 53.8 seconds to load)
427 examples, 2 failures, 5 pending
```

murny:
- The first one I have a fix for in my speed up test branch. It's because on the communities view (if you have no communities and the communites instance variable is nil) it explodes when you try to slice it. So a simple if nil check fixes that.
- Second I have never seen? That's interesting?

mbarnett:
For that second one, it might be worth checking if Rails is crashing on that page. If Rails crashes, it will just run the match against the page it gets back, usually a better_errors page -- this is either better_errors being dumb and not returning a 500 status, or Rspec being dumb and not doing something sane with that status if you don't explicitly test for a 200 OK before running a matcher.

pgwillia:
@murny would it be okay to add your fix for the collection_spec test to this PR? Weird that current master all tests pass but this one fails consistently for this branch.  Don't see anything that should cause this behaviour to change?!

murny:
From looking at the specs it should be failing with the current code? (i'm suprised its not in master?)

The spec before the failing one(line 493) is basically create a collection, make it a community, visit community index page, see if it's rendered on that page. Which passes. Then the next one and the one causing issues (line 507) is edit that collection, make it no longer a community, visiit community page, see if it's not rendered. This should fail with the current page as if the `@user_communities` is nil, the communities index page will 500. You can't `each_splice` on a nil object. 

I think matt is exactly right. better_errors (and another reason why this shouldn't be in test environment!!!)  is eating up the 500, and our integration test is doing a very poor check of simply just checking if our collection title is on the page or not (which it is not on the better errors page so it assumes it has passed).

Heres all you need to stop this page from 500ing add a if nil check in the view (maybe have an ```@user_communities ||= []``` in the controller where this instance variable is being set would be a smarter approach):

```<% left,right = @user_communities.each_slice( (@user_communities.size/2.0).round ).to_a  if @user_communities.present? %>```

**Comment by mbarnett** [comment]
sooo, I just raised an unconditional error in ApplicationController, verified that it's serving the better errors interface, aaaand:

```
curl -I http://localhost:3000
HTTP/1.1 500 Internal Server Error 
Content-Type: text/plain; charset=utf-8
X-Request-Id: a21e1dbc-a63c-4003-b060-680aeff91d47
X-Runtime: 0.023557
Server: WEBrick/1.3.1 (Ruby/2.2.1/2015-02-26)
Date: Wed, 23 Nov 2016 17:14:07 GMT
Content-Length: 7232
Connection: Keep-Alive
```

Safari shows the same thing in browser. Can you pry into that test and see if it's really not 500ing there?

**Comment by mbarnett** [comment]
I think the problem is you need to scatter "expect status to be 200" or equivalent _everywhere_, because RSpec's API isn't smart enough to assume that if you run a matcher against a page, 95% of the time you didn't mean "match against the error page"

**Comment by murny** [comment]
Yep this is what's happening. Better_Errors is the issue.

From master branch I put a ` save_and_open_page` after the `visit "/communities"` and this is what is rendered:
image
We then check if `expect(page).not_to have_content("Test Community")`, which is true!

**Comment by mbarnett** [comment]
But, again, the REAL issue is that we're not checking the status code after a visit(). A regular error page that happened to have the word community in the footer would run into the same issue. We need to start checking status codes.

**Comment by mbarnett** [comment]
Mostly what I'm trying to get at is that turning better_errors off in the test environment only papers over the issue -- this would start working, but if some other test in the future happens to match some text on an error page, same "mysterious" issue. 

```expect(response).to have_http_status(200)```

is our only saviour in this troubled world

**Comment by pgwillia** [comment]
I think you guys are right that spec/features/collection/collection_spec.rb:507 has been giving false positives all along. @mbarnett beat me to the proof.

Checking status codes is a good note to add to https://github.com/ualbertalib/Hydranorth2/issues/83. I'd be pleased if @murny's test cleanup contained something like that but otherwise I'm guessing that this isn't high enough priority to get any more time? We should ask at the next backlog grooming or similar meeting.

**Comment by murny** [comment]
Absolutely! It's a poor test no arguing with that. Checking the status code is a bit low level for an integration test. Maybe checking something like one of the following:
 ```ruby
expect(current_path).to eq(communities_path)
expect(page).to have_title "ERA | Communities" # We don't replace titles in ERA from the look of it...but still an option
expect(page).to have_css "h1", text: "Communities"
```

mbarnett:
Yeah, I'd add a status check here to get it working, but cleaning up EVERY test that isn't checking status is definitely way out of scope

murny:
Thanks for merging this! There's still questions in the main description that I think should be talked about. Mainly around a few gems if they are needed and the whole `spec-views` directory (standalone selenium test suite we have)

pgwillia:
No problem. I'm opening a new issue with the outstanding things in your description. Feel free to add/remove.

Issue #778, opened 2015-10-16
Speed up tests


mbarnett:
Currently the rspec test suite takes upwards of 2 hours to run. Even disabling the slowest test (https://github.com/ualbertalib/HydraNorth/blob/master/db/performance_data.rb), the test suite takes approximately 26 minutes to run.

[Slow tests are an anti-pattern](http://randycoulman.com/blog/2015/07/21/getting-testy-anti-patterns/). This is a big anchor on our overall productivity, as things slow way down any time we need to verify that a change hasn't unintentionally broken something else in the system. 

There are a lot of avenues we can investigate to speed things up:
- [ ] replace the performance test with a mock that simply aims to detect whether the index is performing an N+1 query of files, rather than actually importing 1100+ files into fedora and then timing the render, which is an indirect way of detecting the above, and which takes over an hour.
- [ ] eliminate many of the downloads. At least one spec (possibly redirect_spec, although I haven't checked) downloads a number of files, one of which is over 100MB in size. It should be possible to side load these through ansible, or commit sanitized data to GitHub, rather than curl 20-30 files over the network during each test run. Additionally, many of these downloads seem to be the same file, several times
- [ ] switch to FactoryGirl#build_stubbed wherever possible to avoid hitting the database whenever it can be avoided.
- [ ] take the above even further and [avoid FactoryGirl altogether](https://robots.thoughtbot.com/speed-up-tests-by-selectively-avoiding-factory-girl) where persistence isn't needed
- [ ] consolidate examples. We define many `let()` variables per test. We should attempt to share these across as many tests as possible, to reduce redundant creation of model instances, etc. It may be preferable to simply reset some values on an object after a test so that it can be shared with another test, rather than clean it out of the database and then create a new identical object in the database.
- [ ] reduce as many calls to `cleanup_jetty` as possible, to eliminate expensive calls to Fedora. Running it after each test may be overkill if the test didn't do anything that will impact other tests.
- [ ] mock interactions with Fedora wherever possible, to eliminate expensive calls to Fedora
- [ ] aim to shorten or eliminate any calls to `sleep()`.  [generic_file_new_form_spec](https://github.com/ualbertalib/HydraNorth/blob/master/spec/forms/generic_file_new_form_spec.rb) spends 90 seconds in `sleep`, for instance
- [ ] [defer garbage collection](http://makandracards.com/makandra/950-speed-up-rspec-by-deferring-garbage-collection)
[...]

PR #1316, opened 2016-11-23
778 speed up tests

murny:
This pull request contains a number of refactors for our test suite. The main goal here is for speed improvements. Previously a full test suite run using master was taking over > 25-30 minutes on my local machine. Here are some timings I got from today on master branch on 427 examples:
- 33 minutes and 46 seconds
- 36 minutes and 46 seconds
- 31 minutes and 29 seconds

After this pull request, it is around ~16 minutes on my local. Here are some timing I have gotten on this PR on 335 examples:
- 16 minutes and 24 seconds
- 15 minutes and 48 seconds
- 15 minutes and 26 seconds
- 15 minutes and 37 seconds
- 16 minutes and 46 seconds

After merging this with master (which has my code cleanup commits) I am getting around ~11 mins, Here are some timings I have gotten:
- 11 minutes 33 seconds
- 11 minutes 29 seconds
- 11 minutes 30 seconds
- 11 minutes 33 seconds

Root cause? Seems to be `better_errors` being in the test group and running in our tests! Nice 4 minute speedup! Interesting.

I assume this will be a vast speed up on our VM's as well.

I only really looked deeply into the controller tests and refactored those, so there is definitely more improvements that could be made especially for the integration/feature test which consumes over half the total speed time! (Most of the folders (controllers/models/jobs/etc) take around 1-2 mins each, but `/features` consumes almost 9-10 minutes of the total time)

Ultimately the biggest bottleneck is fedora. So doing this smarter will net the biggest gains. Simply using a `GenericFile.new` instead of a `GenericFile.create` can net an easy 10 seconds! So rule of thumb is don't create an object when `build`, `build_stubbed`, `attributes_for`, `spy`, or `double` will do. Database persistence is slow (especially in fedora)! We might want to start mocking more?

On the fedora note, I think we should rethink on how we are using FactoryGirl, if at all for fedora. FactoryGirl only works well when you are using transactions. For a normal database, FactoryGirl will create the object in a transaction, run the example, then rollback the transaction. So it will always cleanup after itself if everything is in a transaction. If you use FactoryGirl in an `before(:all)` you will notice it will dirty the database as it's no longer in the transaction for an example. Fedora is the same way, it does not support transactions. So it will always create objects but never clean them up. So if you have something like the following:

```ruby
let!(:file) do
  FactoryGirl.create(:generic_file)
end
```
 And you had 9 examples in your spec. It will leave 9 objects behind. This is not ideal at all and not the expected behaviour when using FactoryGirl! This is big reason we have `cleanup_jetty` or `GenericFile.delete_all`, throughout the test suite in the before and after blocks which is a pretty big code smell. 

In regards to cleaning up your data, always use the `cleanup_jetty` method which is just calling `ActiveFedora::Cleaner.clean!`. Running some benchmarks on a typical spec you will see cleanup_jetty is much much faster then other cleanup methods. Which seems weird? Maybe someone with some knowledge of fedora can explain this?

Say you are creating 3 Generic files in a before each block:

```ruby
before do
  @gf1 = FactoryGirl.create(:genericfile)
  @gf2 = FactoryGirl.create(:genericfile)
  @gf3 = FactoryGirl.create(:genericfile)
end
```
And then you might clean up these objects in an after each block,
```ruby
after do
  # Could just called GenericFile.delete_all or GenericFile.destroy_all
  # Could just call cleanup_jetty
  @gf1.delete
  @gf2.delete
  @gf3.delete
end
```
On a typical spec using `.delete` on each object  I was getting ~6 seconds for the above (6.13, 6.25 and 5.96). Using `GenericFile.delete_all` I was getting ~8 seconds (7.82, 7.7, 8.29). And finally using `cleanup_jetty` I was getting ~5 seconds (4.55, 4.53, 5.23). So lesson is, use `cleanup_jetty` where possible.

The other big issues outside of fedora is just smarter testing in general. We are testing the wrong things (tests to make sure underlining gems, ruby/rails, or rspec are working correctly shouldn't be needed),  have useless test setup's, lot's of copy pasta, etc. I hardly touched the big bottleneck (fedora) and simply just refactoring our tests to be smarter shaved off alot of minutes on our total test time! It's important that we make an effort to write tests that are accurate and effective as well as fast.

Other things to look out for in the future:

- data stores (such as database, job queues, ActionMailer::Base.deliveries, etc) should return to a consistent state between each spec, regardless of the order specs are run in.
  - cleaning database before and after every example shouldn't be needed? (via database_cleaner or cleanup_jetty)
  - tests should be configured to run with seeds/random order should help with this
- Use cleanup_jetty vs delete/delete_all/destroy_all
- Don't mark a feature as requiring JavaScript (through js: true in RSpec) unless it's actually required for the test to be valid. Headless browser testing is slow!
- Should be using `rails_helper.rb` for rails code  and `spec_helper.rb` (light weight version) for non rails code

Code style things:
- don't use `.tap` on factorygirl/activerecord/activefedora objects (let's not use blocks in general? If you are using blocks, `.new/.build` then call save in the block! Do not use `.create` and then call `.save` inside the block! We seem to abusing them because we use `#apply_depositor_metadata` helper method all over the place. Which from my understand just sets the `depositor_id` and `edit_users` with user's email?  http://www.rubydoc.info/gems/hydra-access-controls/10.0.1/Hydra%2FWithDepositor%3Aapply_depositor_metadata)
- Try not to use `subject` in rspec, especially implicit
- Don't use one liner `it {}` syntax if not in isolation
- Pick either `before` or `before(:each)` syntax and use everywhere. `:each` is the default. I would prefer to be specific about it, as I've seen a few cases where we think it is a `before(:all)`. Regardless be consistent about it.

Matt makes some great recommendations in the original issue as well: https://github.com/ualbertalib/HydraNorth/issues/778

Some resources from around the web:
- http://betterspecs.org/
- Thoughtbot's blog has a ton of great testing articles: 
  - https://robots.thoughtbot.com/speed-up-tests-by-selectively-avoiding-factory-girl
  - https://robots.thoughtbot.com/four-phase-test
  - etc.
- https://github.com/thoughtbot/guides/tree/master/style/testing
- https://gitlab.com/gitlab-org/gitlab-ce/blob/master/doc/development/testing.md
- https://github.com/eliotsykes/rspec-rails-examples


lib/tasks/dataverse_migration.rake
```diff
@@ -54,33 +55,33 @@ namespace :migration do
       MigrationLogger.info "Processing the object #{object_id}"
       #reading the metadata file
       metadata_file = Nokogiri::XML(File.open(file))
-      metadata = metadata_file.xpath("//ddi_to_dcterms",NS)
+      metadata = metadata_file.xpath("//ddi_to_dcterms", MigrationConstants::NS)
```
murny: Noticed some weird dependencies in our rake tasks. We are depending on this NS constant that is defined in the migration.rake task. If it is a dependency and needed in other rake tasks, then make it so. Never make the assumption it is just loaded and avaiable across rake tasks. (When testing in isolation this was a big pain figuring this out). This is what the next few files are doing (lots of white space diffs). I pulled the NS constant into a MigrationConstants module and requiring it where needed.

spec/actors/hydranorth/generic_files/actor_spec.rb
```diff
@@ -1,12 +1,10 @@
 require 'spec_helper'
 
 describe Hydranorth::GenericFile::Actor do
-  include ActionDispatch::TestProcess
-  
-  let(:user) { FactoryGirl.find_or_create(:user) }
-  let(:generic_file) { FactoryGirl.create(:generic_file) }
+  let(:user) { FactoryGirl.build(:user) }
+  let(:generic_file) { FactoryGirl.build(:generic_file) }
```
murny: Here's a great example of showcasing the bottlenecks of Fedora. In this particular test, we do not need a creation of a GF, building the GF in memory will do just fine. This test went from taking 10 seconds to < 1 second simply because of this small change. Lesson? Try not to persist to the Fedora if we do not need to.

spec/controllers/admin/become_controller_spec.rb
```diff
@@ -1,41 +1,41 @@
...
+  let(:user) { FactoryGirl.create(:jill) }
+  let(:other_user) { FactoryGirl.create(:alice) }
+
   context 'admin' do
     before do
       sign_in admin
     end
 
     it 'should become user' do
-      get :index, id: user.email, use_route: :admin
```
murny: Controller specs are unit test. They should have zero dependency on how the routes are working. I removed everywhere we are using `use_route`

spec/controllers/catalog_controller_spec.rb
```diff
@@ -1,8 +1,6 @@
 require 'spec_helper'
 
 describe CatalogController, type: :controller do
-  routes { Rails.application.class.routes }
```
murny: This should be the default case. Shouldn't need to specify this in every controller test

spec/controllers/catalog_controller_spec.rb
```diff
@@ -206,12 +200,8 @@ def year_created_facet
     end
 
     describe "year_created facet search" do
-
-      before do
```
murny: We use before hooks in alot of cases where it is only being used for one example. Why? Before blocks are meant for drying up duplication across multiple examples.

spec/controllers/communities_controller_spec.rb
```diff
@@ -1,32 +1,8 @@
 require 'spec_helper'
 
 describe CommunitiesController, :type => :controller do
-  routes { Rails.application.class.routes }
```
murny: Yeah....Have no idea whats going on here? We are doing a ton of setup only to do a smoke test on the controller. Let's remove this useless setup.

spec/controllers/legacy_users_controller_spec.rb
```diff
@@ -1,17 +0,0 @@
-require 'spec_helper'
-
-describe UsersController, :type => :controller do
```
murny: We are really bad at managing specs in our test suite. We have three seperate users controller tests, which shouldn't be the case. I consolidated them into one users_controller_spec.rb.

We are testing the wrong things in certain tests, we have like integration tests in /tasks, /forms, etc, we aren't naming the test files the correct way, etc etc.

spec/controllers/redirect_controller_spec.rb
```diff
@@ -1,15 +1,11 @@
 require 'spec_helper'
-require 'rake'
```
murny: Bad copy paste? A controller spec shouldnt depend on rake/fileutils

spec/controllers/selects_collections_controller_spec.rb
```diff
@@ -8,75 +10,74 @@ class SelectsCollectionsController < ApplicationController
...
-      end
-      it "should not return public collections" do
-        @user_communities = subject.find_communities
-        expect(@user_communities.map(&:id)).not_to include @collection.id
+      it "should return public communities and not public collections" do
+        expect(subject.find_communities.map(&:id)).to match_array [@community.id, @no_edit_community.id]
       end
     end
 
     describe "Regular User Read Access" do
```
murny: Previously the Read Access and Edit Access were the exact same? I got no idea what this code is doing but I assume its not testing the right things. Just passing in `:read` and `:edit` respectfully and all seems to still pass?

spec/controllers/selects_collections_controller_spec.rb
```diff
@@ -8,75 +10,74 @@ class SelectsCollectionsController < ApplicationController
...
       describe "not signed in" do
         it "should error if the user is not signed in" do
-          expect { subject.find_communities([:read, :edit]) }.to raise_error
+          expect { subject.find_communities([:read]) }.to raise_error(UncaughtThrowError, 'uncaught throw :warden')
```
murny: raise_error is very dangerous. It is a general catch all which results in false positives. If we are using it, let's be specific on what type of error we are expecting to happen.

spec/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,36 +1,28 @@
...
-      :provider => 'shibboleth',
-      :eppn => 'myself@testshib.org',
-      :uid => 'myself@testshib.org'
-    })
-
-  end
-
+describe Users::OmniauthCallbacksController, :type => :controller do
   describe "GET shibboleth" do
-    let(:action) { :shibboleth }
```
murny: useless let? Never being referenced more then once.

spec/controllers/redirect_controller_spec.rb
```diff
@@ -20,7 +16,7 @@
   describe "#item" do
     it "redirects to item page" do
       get :item, uuid: fedora3uuid1
-      expect(response).to redirect_to "http://test.host/files/#{gf.id}"
+      expect(response).to redirect_to generic_file_path(gf.id)
```
murny: I much rather be using the route helpers instead of hard coding these? (what happens if these route changes? now someone has to go update these tests)

spec/controllers/single_use_links_viewer_controller_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'spec_helper'
 
+# TODO: Useless test?
```
murny: This test is kinda pointless. Its basically checking if ruby assignment operator (=) works.
Controller Code:
```ruby
self.presenter_class = Hydranorth::GenericFilePresenter
```
Test Code:
```ruby
 expect(SingleUseLinksViewerController.presenter_class).to be Hydranorth::GenericFilePresenter 
```

spec/factories/generic_files.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
```
murny: Let's not use Sufia's generic file factory which we have no control over and it's not even valid in our code base (linter fails on it). Here's a simpler version of theirs:

spec/factories/users.rb
```diff
@@ -1,5 +1,6 @@
-require 'bcrypt'
+require 'bcrypt' # We don't have this in our gemfile
 
+# We have way to many different users here. Should have a few test users depending on roles? User, Admin, etc.
```
murny: As comment says, we don't need all these factories... Needs to be way more simpler then this. Lets name them smarter (`jill` doesn't explain what the user can do) and you only need many 3-4 factories for the different user permission levels (have maybe two for a register user, an admin user, and maybe like ccid/legacy user?)

spec/controllers/redirect_controller_spec.rb
cwant: One small thought about this (and please excuse any newbie ignorance):  why not take it one step further and use generic_file_path(gf) (excluding the id method). Then the test becomes immune to "to_param" being overridden in the model.

spec/factories/users.rb
```diff
@@ -14,6 +15,7 @@
       email 'aliceuser@example.com'
     end
 
+    # TODO: Bad factory, we are already seeding the test database with this user? Potential conflict with email being taken already
```
murny: We have a seeds file that has this user in already. Could be potential issue if we weren't wiping the database on every test.

spec/features/admin_tasks_spec.rb
```diff
@@ -80,23 +81,12 @@
 
 end
 
-def setup(user)
```
murny: Never used.

spec/features/admin_tasks_spec.rb
```diff
@@ -80,23 +81,12 @@
...
-      visit "/dashboard/files"
-end
-
 def create_batch(file)
   within("#document_#{file.id}") do
     check "batch_document_#{file.id}"
   end
 end
 
-def search(query="")
```
murny: Only used once, lets just inline it

spec/features/admin_delete_spec.rb
```diff
@@ -49,12 +51,7 @@
       expect(page).to_not have_selector(:link_or_button, 'Delete Selected')
     end
 
-  end 
-end
-
-def setup(user)
```
murny: Only used twice? Don't need to abstract this I'd argue

spec/features/agreement_spec.rb
```diff
@@ -1,14 +1,12 @@
 require 'spec_helper'
 
-
 describe "Display agreement", :type => :feature do
-  before do
-    sign_in :admin
```
murny: This is important! You should always use the lowest amount of authorisation as possible. Do not test functionality as an admin when a basic user or no user can suffice. Heck, do you even need to be logged in to see the agreement page now that I think about it? Answer is No.

spec/features/collection/collection_spec.rb
```diff
@@ -491,8 +491,6 @@
     end
 
     it "should set Official and Community flags" do
-      visit "/collections/#{community.id}/edit"
```
murny: This is duplicated. We are already visiting this in a before(:each)

spec/helpers/generic_file_helper_spec.rb
```diff
@@ -1,17 +1,17 @@
 require 'spec_helper'
 
-describe 'download link', :type => :feature do
+describe GenericFileHelper, type: :helper do
```
murny: Bad copy paste? Also the file name should have helper in it.

spec/models/ability_spec.rb
```diff
@@ -66,152 +66,142 @@
...
+  after(:all) do
     cleanup_jetty
   end
 
   describe "a user with no roles" do
-    let(:guest) { nil }
-    subject { Ability.new(guest) }
-
-    it { is_expected.not_to be_able_to(:create, GenericFile) }
-    it { is_expected.not_to be_able_to(:edit, file) }
```
murny: After this refactor this spec went from ~90 seconds to <20 seconds. Why? Because we are using the one line `it {}` syntax throughout this file. This should only ever be used when your code is in isolation. If it is not (ones that integrate with a DB, an external webservice, or end-to-end-tests). Then you should not use this. I would argue you shouldn't use it anways. Having it in a it block is much more expressive and easier to understand. The big issue here is the before/after blocks are ran on every one of these and there was like 80+ in this spec which makes this super slow.

spec/models/ability_spec.rb
```diff
@@ -66,152 +66,142 @@
...
+      expect(guest_abilities).not_to be_able_to(:download, restricted_file)
+      expect(guest_abilities).not_to be_able_to(:download, restricted_doc)
+      expect(guest_abilities).to be_able_to(:read, institutionally_restricted_file)
+      expect(guest_abilities).not_to be_able_to(:download, institutionally_restricted_file)
+      expect(guest_abilities).not_to be_able_to(:download, institutionally_restricted_doc)
+    end
   end
 
   describe "a registered user" do
-    subject { Ability.new(user) }
```
murny: I disagree with subject.

- implicit subject should never be used. (where rspec makes assumptions on the subject and bases it off an instantiation of the described class): Even the rspec docs will say this: https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/subject/implicitly-defined-subject
- Explicit subject is better but you should be naming the subject. Rspec docs will tell you to do this as well: https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/subject/explicit-subject.

 But in this case I would argue....why not just use a let block instead? Which is what I did here

spec/controllers/redirect_controller_spec.rb
murny: I can agree with that! Good call!

Gemfile
```diff
@@ -75,23 +75,22 @@ gem 'rest-client'
...
 
 group :test do
-  gem "rspec-its"
+  gem "capybara"
+  gem "capybara-select2"
+  gem "database_cleaner"
   gem "ci_reporter_rspec"
+  gem "poltergeist", "~> 1.5"
+  gem "show_me_the_cookies"
   gem "timecop"
```
murny: As of Rails 4.1+, timecop isn't really needed (unless your using some it's advance features for accelerating time,etc.). Rails comes with TimeHelpers for mocking out/freezing/jumping with time.

More Info on TimeHelpers: http://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html

So to freeze time and unfreeze time you will just do the following:
```ruby
      before do
        # Freeze time
        travel_to Time.zone.now
      end

      after do
        # Unfreeze time
        travel_back
      end
```

Gemfile
```diff
@@ -75,23 +75,22 @@ gem 'rest-client'
 gem 'noid', '~> 0.8'
 
 group :test do
-  gem "rspec-its"
```
murny: We were using rspec-its to backport rspec's `its {}` functionality which was removed from rspec. We were using this in three places. I have refactored these so we no longer need this. `its {}` syntax does lots of magic and isn't intuitive to understand. Reason it has been removed from rspec. More info about why it was removed here:
https://gist.github.com/myronmarston/4503509

Gemfile
```diff
@@ -75,23 +75,22 @@ gem 'rest-client'
...
   gem "ruby-debug-passenger"
-  gem "selenium-webdriver"
+  gem "selenium-webdriver" # used spec-views (legacy?)
   gem "jettywrapper"
-  gem "capybara"
-  gem "poltergeist", "~> 1.5"
   gem "factory_girl_rails"
-  gem "database_cleaner"
-  gem "capybara-select2"
-  gem "show_me_the_cookies"
```
murny: Can't see any reason why most of these test gems are in the development group as well? Some of them are valid (factory_girl/rspec-rails has generators) but these other ones I moved up to the test group.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -20,21 +18,20 @@ def year_created_facet
...
 
-    GenericFile.create(title: ['Test 2 Document'], filename: ['test2.doc'], contributor: ['Contrib2'], read_groups:['public']).tap do |f|
+    GenericFile.create(title: ['Test 2 Document'], filename: ['test2.doc'], contributor: ['Contrib2'], read_groups:['public']) do |f|
       f.apply_depositor_metadata('qw1')
       f.save
-      @gf2 = f.id
     end
 
-    GenericFile.create.tap do |f|
+    GenericFile.create do |f|
```
murny: Two things about how we are creating our records that I see all across our test suite:
- You do not need `.tap`! ActiveRecord (and it appears ActiveFedora) are block friendly already, so you can just pass a block to it without this
- We should **NOT** be doing using blocks this way anyways. Especially with using a  `.create` (`.new` is okay I guess)! Why? 

Because this actually does two transactions to the database. One for the creation and one for the updating. In ActiveRecord the following code would behave like so:

```ruby
User.create do |u|
     u.name = 'example'
     u.save
end
```
```
(0.2ms)  begin transaction
SQL (0.4ms)  INSERT INTO "users" ("created_at", "updated_at") VALUES (?, ?)  [["created_at", 2016-11-24 21:28:19 UTC], ["updated_at", 2016-11-24 21:28:19 UTC]]
(68.1ms)  commit transaction

(0.1ms)  begin transaction
SQL (0.4ms)  UPDATE "users" SET "name" = ? WHERE "users"."id" = ?  [["name", "example"], ["id", 2]]
(61.5ms)  commit transaction

=> #<User id: 2, name: "example", created_at: "2016-11-24 21:28:19", updated_at: "2016-11-24 21:28:20"> 
```

I am pretty confident it works just like this in ActiveFedora as well (if not worst). And considering how slow fedora is, this is a very big performance issue!

So let's do things smarter. If we are using blocks, use `.new`. If not just pass the attributes into the create method instead (Such as `User.create(name: 'example')`).

spec/controllers/my/all_controller_spec.rb
```diff
@@ -2,80 +2,59 @@
...
-
-  describe "User" do
-    before do
-      sign_in user
+    let!(:unrelated_file) do
+      FactoryGirl.create(:generic_file, depositor: other_user)
     end
 
-    it 'should respond with error' do
-      expect{ get :index, user_route: dashboard }.to raise_error
```
murny: This test isn't testing the right thing. A basic user can visit their /my/all_controller index, you do not need to be an admin to do this.

But most importantly this shows an example of a false positive from using raise_error. This is only passing because we miss spelled `use_route` as `user_route` which isn't valid, and we are capturing this error instead!

spec/spec_helper.rb
```diff
@@ -130,29 +123,7 @@ def main_app
 
   config.include Devise::TestHelpers, type: :controller
   config.include Devise::TestHelpers, type: :helper
-  config.include EngineRoutes, type: :controller
   config.include Warden::Test::Helpers, type: :feature
   config.after(:each, type: :feature) { Warden.test_reset! }
   config.infer_spec_type_from_file_location!
 end
-module FactoryGirl
-  def self.find_or_create(handle, by=:email)
```
murny: This is a weird addition to FactoryGirl. Would argue you shouldn't be using factory girl to lookup a object from the database? Especially if we are clearing out the database after every example. Feel like you should have a bit of understanding of what your trying to do instead of just calling `FactoryGirl.find_or_create()` everywhere in our test suite. Can we get away with a build? a build_stubbed? Then we should use create?

spec/spec_helper.rb
```diff
@@ -5,13 +5,13 @@
 ENV["RAILS_ENV"] ||= 'test'
 require File.expand_path("../../config/environment", __FILE__)
 require 'rspec/rails'
-require 'rspec/its'
 require 'capybara/rails'
 require 'capybara/rspec'
 require 'capybara/poltergeist'
-require 'factory_girl_rails'
-require File.join(Sufia::Engine.root, 'spec/factories/generic_files')
-require File.join(Sufia::Engine.root, 'spec/support/features')
```
murny: This pulls in a bunch of helpers from sufia which we are only using the cleaner for cleanup_jetty. Which is a custom implementation. Let's remove this and create our own cleanup_jetty (I am using ActiveFedora::Cleaner.clean! to do this instead).

spec/support/factory_girl.rb
```diff
@@ -0,0 +1,15 @@
+require 'factory_girl'
+
+RSpec.configure do |config|
+  config.include FactoryGirl::Syntax::Methods
```
murny: Brought this in, this allows you to use create/build instead of namespacing them like FactoryGirl.create/FactoryGirl.build.

Also it's a good practice to lint your factories, see commented code.

spec/support/rake_helper.rb
```diff
@@ -0,0 +1,17 @@
+require 'rake'
+
```
murny: One way of doing this differently instead of loading your rake tasks in isolatation is you can just load all them in before your test suite? like:

```ruby
RSpec.configure do |config|
  
  config.before(:suite) do
    Rails.application.load_tasks
  end

end
```

But regardless the issue with the way we were previously doing this was we were constantly reloading the rake files over and over again resulting in a ton of spam to the logs for overwriting constatns that were already defined.

spec/support/search_helper.rb
```diff
@@ -1,8 +1,8 @@
 module SearchHelper
```
murny: Moved this into support directory instead

spec/support/session_helpers.rb
```diff
@@ -0,0 +1,17 @@
+module Features
+  module SessionHelpers
+    def sign_in(user)
```
murny: Extracted this out into support helper. I would say we should name this differently? Like `login_as` or something different as the devise helpers (which are only being loaded currently in the controller/helper tests) also use `sign_in` name

spec/tasks/migration_rake_spec.rb
```diff
@@ -1,484 +1,436 @@
...
-    subject { @file }
-
     it "visibility should have been updated" do
-      expect(subject.institutional_visibility?).to be true
-      expect(subject.read_groups).to include Hydranorth::AccessControls::InstitutionalVisibility::UNIVERSITY_OF_ALBERTA
-      expect(subject.read_groups).to include Hydra::AccessControls::AccessRight::PERMISSION_TEXT_VALUE_PUBLIC
+      run_rake_task('migration:update_ccid_visiblity', 'spec/fixtures/migration/uuids_ccid_visibility')
+      @generic_file.reload
+      expect(@generic_file.institutional_visibility?).to be true
+      expect(@generic_file.read_groups).to eq [Hydra::AccessControls::AccessRight::PERMISSION_TEXT_VALUE_PUBLIC, Hydranorth::AccessControls::InstitutionalVisibility::UNIVERSITY_OF_ALBERTA]
```
murny: I see we do alot of individual checking of arrays using the `to include <value>`. Why not just check if the array is exactly how it should be? 

Probably shouldn't be using eq matcher either. Should be using `match_array` matcher instead. Will fix this.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -54,14 +51,11 @@ def year_created_facet
...
       f.save
-      @gf3 = f.id
     end
   end
 
   after :all do
-    GenericFile.find(@gf1).delete
-    GenericFile.find(@gf2).delete
-    GenericFile.find(@gf3).delete
+    cleanup_jetty
```
murny: See comment in main issue: https://github.com/ualbertalib/HydraNorth/pull/1316#issue-191396103

`cleanup_jetty` is far faster then using individual deletes or a delete all. (We are looking up the GF here with a find first, but even doing this smarter by making @gf1 to be the actual record and not the id, and calling delete on the record it is still far slower)

spec/controllers/my/files_controller_spec.rb
```diff
@@ -1,44 +1,39 @@
...
+      it "should respond with success and shows the correct documents" do
         get :index
         expect(response).to be_successful
-      end
-
-      it "shows the correct documents" do
-        get :index
+        document_list_ids = assigns[:document_list].map(&:id)
         # shows documents shared with me
-        expect(assigns[:document_list].map(&:id)).to include(@read_shared_with_me.id)
```
murny: Looping over this object everytime to create an array of ID's could be slow? Lets do it once and assign it to a variable, and then use this variable in the asserts?

spec/features/collection/collection_spec.rb
```diff
@@ -258,8 +258,7 @@
       check "batch_document_#{collection.id}"
       click_button 'Add to Collection'
       expect(page).to have_content("Select the collection to add your files to:")
-      page.execute_script("document.getElementById('id_" + community.id + "').checked = true")
-      expect(find_field("id_#{community.id}")).to be_checked
+      choose "id_#{community.id}"
```
murny: This is a radio button, let's use the actual capybara methods and not hack the form with JS?

spec/features/collection/collection_spec.rb
```diff
@@ -287,15 +286,13 @@
   it { expect { visit "/collections/#{collection.id}" }.to_not raise_error }
 
   describe 'paginate collections' do
-    let!(:collection_delete) do
```
murny: More bad copy pastey?

Why create a let block here if you are never using the variable? Also this has nothing to do with collection_delete and was copied and pasted here I assume.

spec/features/community_spec.rb
```diff
@@ -50,25 +50,20 @@
...
     end
 
-    it 'should render quickly' do
-      expect( Benchmark.realtime { visit "/communities" }).to be < 5 
-    end
-
-
     it 'should have all communities and collections' do
-      visit "/communities" 
+      expect( Benchmark.realtime { visit "/communities" }).to be < 5 # should render quickly
```
murny: We should condense our tests more often! Don't need two examples for this when we are just testing the same thing pretty much (especially with lots of setup code! Database IO is slow with fedora!). This tiny change halfs this test speed execution.

spec/features/display_home_page_spec.rb
```diff
@@ -1,24 +1,22 @@
 require 'spec_helper'
 
 describe "Home Page", type: :feature do
-  let(:abby) { FactoryGirl.find_or_create :user_with_fixtures }
-  
-  before do
-    sign_in abby
```
murny: The first example, doesn't need a user to be signed in?  Let's do the bare minimal in our setups.

spec/features/facet.rb
```diff
@@ -2,12 +2,7 @@
 
 describe 'facet', :type => :feature do
 
-  before do
```
murny: We should't need to cleanup before and after a test? Assuming we aren't polluting the database from other specs

spec/features/saml_spec.rb
```diff
@@ -4,8 +4,6 @@
 
   it { expect { visit '/users/sign_in' }.to_not raise_error }
 
-  after { cleanup_jetty }
```
murny: This entire feature is based around User model (aka ActiveRecord/Mysql). It has nothing to do with Fedora. Why are constantly cleaning fedora in this spec?

spec/features/saml_spec.rb
```diff
@@ -128,14 +123,7 @@
   end
 
   describe 'Users with confirmed CCID accounts' do
-    let!(:user) { FactoryGirl.create :user, email: emailize_uid(OmniAuth.config.mock_auth[:shibboleth][:uid]), ccid: OmniAuth.config.mock_auth[:shibboleth][:uid] }
-    after :each do
-      cleanup_jetty
-    end
-
-    after :all do
-      User.destroy_all
```
murny: We are using FactoryGirl....we don't need to clean up all the users? FactoryGirl will roll back the transaction of the user. More unneeded database I/O

spec/models/generic_file_spec.rb
```diff
@@ -1,12 +1,6 @@
 require 'spec_helper'
 
 describe GenericFile, :type => :model do
-  let(:user) { FactoryGirl.find_or_create(:user) }
-
-  before(:each) do
-    @file = GenericFile.new
-    @file.apply_depositor_metadata(user.user_key)
```
murny: All useless setup never being used...removed

spec/models/solr_document_spec.rb
```diff
@@ -24,10 +22,8 @@
   end
 
   describe "dissertant" do
-    before do
-      subject['dissertant_tesim'] = 'Test User'
```
murny: Pointless to have this type of logic in a before block if its only being run with one example. Just put it in the example?

pgwillia:
This is awesome! (So speedy) But why are we seeing a different number of examples? And is this only failing for me?

Failures:

  1) SelectsCollectionsController Select Communities Regular User Read Access not signed in should error if the user is not signed in
     Failure/Error: expect { subject.find_communities([:read]) }.to raise_error(UncaughtThrowError, 'uncaught throw :warden')
     NameError:
       uninitialized constant UncaughtThrowError
     # ./spec/controllers/selects_collections_controller_spec.rb:58:in `block (5 levels) in <top (required)>'

  2) SelectsCollectionsController Select Communities Regular User Edit Access not signed in should error if the user is not signed in
     Failure/Error: expect { subject.find_communities([:edit])  }.to raise_error(UncaughtThrowError, 'uncaught throw :warden')
     NameError:
       uninitialized constant UncaughtThrowError
     # ./spec/controllers/selects_collections_controller_spec.rb:73:in `block (5 levels) in <top (required)>'

Finished in 11 minutes 52 seconds (files took 11.15 seconds to load)
335 examples, 2 failures, 6 pending

mbarnett:
Oh, I had the migration spec set up to skip, on further investigation

What version of ruby are you using? UncaughtThrowError is new in 2.2, I think. 

We should probably update our gemfile to reflect a minimum ruby version.

pgwillia:
That's likely it.  I see that ruby 2.1.5 is what was using when running these specs.

murny:
The different amount of examples is due to refactoring. Lots of places where we were testing the exact same thing but having it as different examples (which means before/after blocks run more times then needed). For example models/ability_spec.rb (https://github.com/ualbertalib/HydraNorth/pull/1316/files#diff-e9501dda6c23fd3e1076e2dff66d0f68), has 84 examples and is now down to under 10 as I have refactor this to get away from the one liner `it{}` syntax which should never be used when you are not in isolation.

mbarnett:
I think she meant different number of tests between her run and mine (I had 10 fewer tests run because I skipped a file).

mbarnett:
Any objections to adding a Ruby 2.2 minimum version declaration?

murny:
If production is using 2.3.3 why not bump it to 2.3 haha? But no, no objections here

PR #1320, opened 2016-11-28
#1313: Fix indexing of DOI field

murny:
Allow searching to be exact on the DOI. So when searching 'doi:10.5072/FK22Z17V70 ', not everything with a DOI is in the search results.

This change makes it so that when you 'doi:10.5072/FK22Z17V70', only an object with a DOI of 'doi:10.5072/FK22Z17V70' will appear in the search result


app/models/concerns/hydranorth/generic_file/metadata.rb
```diff
@@ -81,7 +81,7 @@ module Metadata
         property :remote_resource, predicate: ::UALTerms.remote_resource, multiple:false
 
         property :doi, predicate: ::UALId.doi, multiple: false do |index|
-          index.as :stored_searchable
+          index.as :symbol
```
murny: I tried using  `index.as :symbol, :stored_searchable` as this  what I found in other properties we are making a symbol (hasCollectionID, belongsToCommunity, hasCollection, fedora3uuid, fedora3handle, etc, etc). But this does not get what we want. Matt discovered, 

Symbol definition:
https://github.com/projecthydra/solrizer/blob/master/lib/solrizer/default_descriptors.rb#L35

Stored_searchable definition:
https://github.com/projecthydra/solrizer/blob/master/lib/solrizer/default_descriptors.rb#L80

You will see on line 80 of the above link, it will take anything that is a string/text and cast it into a text_en which doesn't allow us to search and match whole words.

app/models/concerns/hydranorth/generic_file/doi_states.rb
```diff
@@ -54,15 +54,21 @@ def withdraw_doi
         end
 
         def handle_doi_states
-          return unless doi_fields_present?
-          # TODO: handle unminted/unsynced states?
```
murny: New added logic below handles the case of unminted/unsynced states. So if a resque job ever fails and a generic file ends up in one of these states, it's possible to resave the generic file to create a new job to update/create a DOI. Instead of forever banishing them from getting a DOI.

spec/models/generic_file_spec.rb
```diff
@@ -153,7 +153,8 @@
       expect(local[Solrizer.solr_name("mime_type")]).to eq ["image/jpeg"]
       expect(local['all_text_timv']).to eq('abcxyz')
       expect(local[Solrizer.solr_name('belongsToCommunity')]).to eq [community.id]
-      expect(local[Solrizer.solr_name('doi')]).to eq ['doi:10.5072/FKEXAMPLE']
+      expect(local[Solrizer.solr_name('doi', :symbol)]).to eq ['doi:10.5072/FKEXAMPLE']
```
murny: I have zero knowledge about Solr, but just curious why we using this `Solrizer.solr_name` helper to access the solr variables? Why not just do `doi_ssim`? Especially if you must have to specify exactly what your looking for (symbol), as it defaults to :stored_searchable (`_tesim`)? Even above on line 154, you can see it doesn't support all the types. It does not support termVectors. 

Also it seems we duplicated majority of our fields in solr? Most of our fields have both a `*_tesim` and a `*_sim` like so:

```
 "language_tesim"=>["Arabic"],
 "language_sim"=>["Arabic"],
```

One for English text stored index multivalued and one for stored (or is this s for string?) index multivalued? Any reason for this?

pgwillia: I think it's a good idea to use the solr_name helper to hide the configuration details. Those suffixes are all dynamic fields from the Solr schema.xml. Unfortunately, Solrizer isn't exhaustive as you've noticed.

The  reason for having both fields is that one is used for facets and the other is used for search and display. Checking for both of them would just be verifying the catalog_controller, but maybe we should?

```*_sim``` is [string not stored](https://github.com/ualbertalib/HydraNorth/blob/master/solr_conf/conf/schema.xml#L74)

murny: Makes sense. Thanks for the explanation everyone!

PR #1321, opened 2016-11-29
Added a quickstart task.

cwant:
User:   rake hydranorth:quickstart

The batch:ingest_csv seems to be broken (whether I run it through
this task or on it's own.)

### DIFF
--- a/lib/tasks/hydranorth.rake
+++ b/lib/tasks/hydranorth.rake
@@ -261,4 +261,27 @@ namespace :hydranorth do
     end
   end    
 
+  desc 'Quickstart for getting a fresh hydranorth development install ' +
+    'up and running'
+  task :quickstart => [ :environment ] do
+    # Would love to pass these tasks as dependencies, but passing
+    # arguments to dependent tasks is difficult/impossible
+    task('migration:user_migration').
+      invoke('lib/tasks/migration/test-metadata/users.txt')
+    task('migration:era_collection_community').
+      invoke('lib/tasks/migration/test-metadata/community')
+    task('migration:era_collection_community').
+      invoke('lib/tasks/migration/test-metadata/collection')
+    task('migration:era_item').invoke
+    task('hydranorth:update_special_itemtype').invoke
+    task('hydranorth:characterize').invoke
+    task('batch:ingest_csv').
+      invoke('lib/tasks/batch/ERA_batch_ingest.csv',
+             'lib/tasks/batch/files_and_metadata/')
+    task('migration:dataverse_objects').
+      invoke('spec/fixtures/migration/test-metadata/dataverse/')
+    task('sitemap:generate').invoke
+    task('hydranorth:harvest:geonames_cities').invoke
+  end
+
 end


lib/tasks/hydranorth.rake
```diff
@@ -261,4 +261,27 @@ namespace :hydranorth do
     end
   end    
 
+  desc 'Quickstart for getting a fresh hydranorth development install ' +
```
murny: This is more for discussion but what's your opinion on limiting your lines? (I assume this is 80 characters which is the standard in most software communities)?

I personally prefer around 100 (https://github.com/airbnb/ruby/blob/master/rationales.md#line-length) as that's a good size on a modern screen before it start's scrolling. But regardless in this project we have no enforcement on it (no rubocop).  So up to you if you want to concat strings like this. Just curious, be good to know what other's think about this (especially when we nail down these rules in https://github.com/ualbertalib/Hydranorth2/issues/55)?

Also to be nitpicky :stuck_out_tongue_winking_eye: , it should probably be using  a backslash for continuation (`\`)? instead of actual true string concatenation (`+`)?
```ruby
desc 'Quickstart for getting a fresh hydranorth development install '\
 'up and running'
```

murny:
Looks good :+1: ! Good job

murny:
A question for the others, why are we seeding our datastores via multiple rake tasks like this? Why not use like rails `db/seeds.rb`? (or bring in something better like seed_fu if you need extra functionality), especially on mandatory data like some of the collections (which result in 500 errors or broken functionality if they are not seeded)? Something to think about for HydraNorth2

pgwillia:
The migration rake tasks have evolved over time. Their actual original purpose was to move metadata and datastreams from Old ERA (which has been trimmed back to thesisdeposit) to Hydranorth. There was some movement towards using db/seeds.rb to pre-populate the application but found that reusing the migration scripts with some sample data was doing the job and acted as an ongoing smoke test for the migration scripts.

pgwillia:
Great idea!  This is really useful.

I'd like to see a note in the README that brings more attention to this, maybe as part of the 'To Install Application' section.

My other thoughts: 
* I don't regularly run ```batch:ingest_csv``` so it wouldn't surprise me if it is broken. @weiweishi is it something that you're still using?  I'll open an issue to fix or remove it.
* The ```hydranorth:harvest:geonames_cities``` task is very time consuming, doesn't change often and is run by the playbook that builds the hydranorth vagrant box. Not sure if we really want it as part of a 'quickstart' task.

weiweishi:
batch:ingest_csv hasn't been run for a while, but it's still in need for
Leah's batch ingest job, until we find an alternative. Thanks for opening
the ticket. I will examine more closely why the job failed.

We can probably modify the task hydranorth:harvest:geonames_cities to use
different seed files from geonames based on the different environments,
since development environments only really need a small sample data for the
autocomplete function.

Weiwei

*Weiwei Shi*

2-10L Cameron Library
Phone:(780)492-7802
Fax: (780)248-1209
Email: weiwei.shi@ualberta.ca

murny:
I deleted tag 1.7 branch which resulted in closing this issue as this was based off that branch. If still relevant fix this up, rebase it off master and I'll merge it

PR #1322, opened 2016-11-29
uses fcrepo_wrapper to provide Fedora 4.7 to dev environment fixes #1034

### LINKED ISSUE #1034: Upgrade Fedora to current release
Spike to try upgrading Fedora to the current version (4.5) and determine how much work will be needed on the rest of the stack to make it work. Our aim is to catch up to the current Fedora if possible, but if the work involved is too big we may decide to make the Fedora upgrade part of the PCDM migration process (so as to avoid redundant work making our current codebase work with Fedora 4.5 and then throw it all out again when we move on to the PCDM work).

Replaces #293, #465, #466

pgwillia:
* fcr:export was deprecated
 * configured logging
 * updated reset-all and restart-all scripts
 * updated README

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,8 @@ group :development, :test do
   gem "rspec-rails"
   gem "ruby-debug-passenger"
   gem "selenium-webdriver"
-  gem "jettywrapper"
+  gem "jettywrapper", require: false
+  gem "fcrepo_wrapper", require: false
   gem "capybara"
   gem "poltergeist", "~> 1.5"
   gem "factory_girl_rails"
--- a/.fcrepo_wrapper
+++ b/.fcrepo_wrapper
@@ -1,0 +1,7 @@
+port: 8984
+enable_jms: false
+fcrepo_home_dir: tmp/fcrepo4-development-data
+version: 4.7.0
+verbose: true
+java_options:
+ - "-Dlogback.configurationFile=fedora_conf/logback.xml"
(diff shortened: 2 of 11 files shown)


Gemfile
```diff
@@ -61,6 +61,7 @@ group :development, :test do
   gem "ruby-debug-passenger"
   gem "selenium-webdriver"
   gem "jettywrapper"
+  gem "fcrepo_wrapper"
```
murny: Don't know anything about this gem so bare with me,  do these rake tasks ever need to be ran in production environment? Does the wrapper only provide rake tasks/cli? if so you could potentially set it to `required: false` as it doesn't needed to be loaded in every process?

murny:
Looks good :+1: ! Just one small question about the gemfile. Also how does this impact a developer? Do I now have to do `rake jetty:stop && rake fcrepo:stop` and `rake jetty:start && rake fcrepo:start`, to bounce fedora?

Gemfile
pgwillia: These rake tasks never need to be ran in production. It's only rake tasks/cli, I believe. I'll make the change.

pgwillia:
Use ``rake jetty:stop``` and ``rake jetty:start`` to bounce solr.  Use ```rake fcrepo:stop``` and ```rake fcrepo:start``` to bounce fedora.  Could that be clearer in the README?

murny:
That makes sense! Nah README is fine! Just viewed it in it's entirety and makes much more sense now then just the small diff I was seeing.  Thanks!

pbinkley:
When should we merge this? I assume we don't want to be developing on 4.7 until the production environment is ready. Do we want to keep this on a branch until then?

pbinkley:
Obsolete, so closing.

Issue #1319, opened 2016-11-26
Can't edit thesis abstract due to absent graduation date value


leahvanderjagt:
@piyapongch 
I need to edit the abstract of this thesis:

https://era.library.ualberta.ca/files/cb2773v94t

But when I do so, I can't select Fall 2016 (2016-11) from the dropdown menu, and ERA is demanding that I enter that. Can you please look at this?

Thanks, 
Leah

weiweishi:
fixed in 1.7.1

PR #1325, opened 2016-12-05
add a new graduation date to the list 

weiweishi:
address #1319


murny:
Looks good :shipit:!

I will say for HydraNorth2 most of this configuration should be in a MySQL DB as a table and be CRUD-able in the admin UI so Leah or someone else can fix this themselves? Because this looks like a recurring bug/fix that will happen twice every year. For example around 2017-06 Leah is going to make the exact same request. Which isn't reasonable. Maybe this should be an raised as an issue in the hydranorth2 admin gem?

weiweishi:
:+1: on storing the values in DB instead of in the configuration file, many other configuration values in sufia.yml should be set up that way. And :+1:  on including this in the admin UI

pgwillia:
@weiweishi what conditions have to exist to see this value in a drop down?

weiweishi:
Only displays when Admin edit an thesis object. @pgwillia

*Weiwei Shi*

2-10L Cameron Library
Phone:(780)492-7802
Fax: (780)248-1209
Email: weiwei.shi@ualberta.ca

PR #1326, opened 2016-12-05
DOI allow smarter recovery after a fail job happens

murny:
View PR https://github.com/ualbertalib/HydraNorth/pull/1320 first. This is a continuation of work. Mainly look at this commit: https://github.com/ualbertalib/HydraNorth/pull/1326/commits/131f0ff16ee50c72c22cf92e6fa2d014846d019b

This fixes an issue when potentially GenericFiles can become stuck in a state where they can no longer be minted or updated for DOI's. Previously if a job fails (for whatever reason such as bad config, EZID API is down, etc.), then the GenericFile will be stuck in a purgatory state (unminted/awaiting_update). And will never be able to do anything with DOI's unless a developer fixes this bad state. Which isn't ideal. This commit will add the behaviour of rolling back the GF to a usable state if it fails when interacting with EZID.

### DIFF
--- a/app/models/concerns/hydranorth/generic_file/doi_states.rb
+++ b/app/models/concerns/hydranorth/generic_file/doi_states.rb
@@ -7,33 +7,35 @@ module Hydranorth
         after_save :handle_doi_states
         before_destroy :withdraw_doi
 
+        attr_accessor :skip_handle_doi_states
+
         include AASM
 
         aasm do
-          state :unpublished, initial: true
+          state :not_available, initial: true
           state :unminted
           state :excluded
           state :available
-          state :unsynced
+          state :awaiting_update
 
           event :created, after: :queue_create_job do
-            transitions from: :unpublished, to: :unminted
+            transitions from: :not_available, to: :unminted
           end
 
           event :removed do
-            transitions from: :unpublished, to: :excluded
+            transitions from: :not_available, to: :excluded
           end
 
           event :unpublish do
-            transitions from: [:excluded, :unsynced], to: :unpublished
+            transitions from: [:excluded, :awaiting_update, :unminted], to: :not_available
           end
 
           event :synced do
-            transitions from: [:unminted, :unsynced], to: :available
+            transitions from: [:unminted, :awaiting_update], to: :available
           end
 
           event :altered, after: :queue_update_job do
-            transitions from: [:available, :unpublished], to: :unsynced
+            transitions from: [:available, :not_available], to: :awaiting_update
           end
         end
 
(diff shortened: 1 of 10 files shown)


app/models/concerns/hydranorth/generic_file/doi_states.rb
```diff
@@ -54,16 +56,23 @@ def withdraw_doi
...
-          return unless doi_fields_present?
-          # TODO: handle unminted/unsynced states?
-          if doi.blank? # Never been minted before
-            created!(id) if !private? && unpublished?
-          else
-            # If private, we only care if visibility has been made public
-            # If public, we care if visibility changed to private or doi fields have been changed
-            if (unpublished? && transitioned_from_private?) || (available? && (doi_fields_changed? || transitioned_to_private?))
-              altered!(id)
+          if skip_handle_doi_states.blank?
```
mbarnett: Maybe throw in a quick comment here (or on the attr above) just to explain the motivation behind this and the semantics of the flipflop

lib/hydranorth/doi_service.rb
```diff
@@ -28,25 +28,44 @@ def initialize(generic_file)
...
+            if @generic_file.private?
+              @generic_file.unpublish!
+            else
+              @generic_file.synced!
+            end
+            ezid_identifer
+          end
+        rescue Exception => e
+          # EZID API call has failed so roll back to previous state change
+          @generic_file.skip_handle_doi_states = true
```
mbarnett: Maybe add a quick note here that we're skipping this to avoid spinlocking on job creation if the API is down

lib/hydranorth/doi_service.rb
```diff
@@ -28,25 +28,44 @@ def initialize(generic_file)
...
+        begin
+          ezid_identifer = Ezid::Identifier.mint(Ezid::Client.config.default_shoulder, doi_metadata)
+          if ezid_identifer.present?
+            @generic_file.doi = ezid_identifer.id
+            @generic_file.synced!
+            ezid_identifer
+          end
+        rescue Exception => e
+          # EZID API call has failed so roll back to previous state change
+          @generic_file.skip_handle_doi_states = true
```
mbarnett: Same note as below

murny:
Added some documentation and rebased (which fixes broken test you seen, which was fixed in PR #1320)

pgwillia:
Does this address the case where the item was ingested without a DOI but we want it to be assigned one?

murny:
Yeah on the test server when peter was creating a DOI, this is what had happen. EZID wasn't properly configured and the GenericFile he was using became in this purgatory state, which wouldn't allow him to get a DOI again (after we configured EZID). This fix will allow the GenericFile to rollback to a usuable state if it fails and be able to get re-saved at a later time to reissue a DOI.

Issue #1344, opened 2017-01-20
Remove DOI field from submission form and edit form


sfbetz:
Let's remove the DOI field from the item submission form!
![screen shot 2017-01-20 at 11 10 05 am](https://cloud.githubusercontent.com/assets/12801010/22160104/2e5c38e2-df01-11e6-99d0-773a14cfaf42.png)
I don't think it's in the edit form, but let's double check and remove from there as well, just in case.

murny:
This has been fixed in master (https://github.com/ualbertalib/HydraNorth/pull/1345). Ready for release. (@weiweishi)

weiweishi:
Hot fix release 1.7.3 deployed on production.

PR #1345, opened 2017-01-20
Remove DOI field from extra metadata descriptions when creating a new…

murny:
… deposit

Fixes bug found in https://github.com/ualbertalib/HydraNorth/issues/1344


murny:
Yeah completely agree. Should make sure we do this going forward for hydranorth2

weiweishi:
Yes. Agreed. Whitelist makes much more sense. I will create a ticket in HN2 to keep track. 

I will tag a release for @henryzhang87

PR #1353, opened 2017-02-07
Minor fixes to migration.rake: set coll/comm on command line, handle source/relation

pbinkley:
Two minor fixes to facilitate loading the radioactive records:

- Allow passing the community and collection urls on the command line, which makes loading non-migrated items easier (i.e. which don't correspond to an item in old ERA)
- Handle the dcterms:source and dcterms:relation fields, mapping them to GenericFile.source and .related_url

### DIFF
--- a/lib/tasks/migration.rake
+++ b/lib/tasks/migration.rake
@@ -74,15 +74,17 @@ namespace :migration do
   end
 
   desc "batch migrate generic files from modified ERA FOXML file"
-  task :eraitem, [:dir, :migrate_datastreams] => :environment do |t, args|
+  task :eraitem, [:dir, :migrate_datastreams, :usecomm, :usecoll] => :environment do |t, args|
     args.with_defaults(:migrate_datastreams => "true")
     begin
       MigrationLogger.info "**************START: Migrate ERA objects *******************"
       metadata_dir = args.dir
       migrate_datastreams = args.migrate_datastreams == "true"
-      # Usage: Rake migration:eraitem[<file directory here, path included>,<optional: migrate_datastreams: boolean>]
+      usecomm = args.usecomm
+      usecoll = args.usecoll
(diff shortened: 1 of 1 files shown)


lib/tasks/migration.rake
```diff
@@ -74,15 +74,17 @@ namespace :migration do
   end
 
   desc "batch migrate generic files from modified ERA FOXML file"
-  task :eraitem, [:dir, :migrate_datastreams] => :environment do |t, args|
+  task :eraitem, [:dir, :migrate_datastreams, :usecomm, :usecoll] => :environment do |t, args|
```
murny: Would `usecom`/`usecoll` just be a string containing the noid for the community/collection here? Or are they just booleans? Seems like they contain the noid as you are pushing whatever is passed in as args onto the communities/collection noid array below on lines 470-480ish? 

So a very small nitpick might be to maybe name these args different? The verb `use` implies a boolean, aka yes/no in whether you should use comm/coll etc? Maybye `community_noid`/`collection_noid` might be a better name? Additonally a comment with an example of how to use this rake task wouldn't be that bad of idea either? 
Like `# rake migration:eraitem[code/example/directory/, true, comm_noid-123456, coll_noid-123456]` 

Other then that, this code looks good from a quick glance! Feel free to pull it in whenever you want! :shipit:

murny:
LGTM :+1:

lib/tasks/migration.rake
```diff
@@ -74,15 +74,17 @@ namespace :migration do
...
     args.with_defaults(:migrate_datastreams => "true")
     begin
       MigrationLogger.info "**************START: Migrate ERA objects *******************"
       metadata_dir = args.dir
       migrate_datastreams = args.migrate_datastreams == "true"
-      # Usage: Rake migration:eraitem[<file directory here, path included>,<optional: migrate_datastreams: boolean>]
+      usecomm = args.usecomm
+      usecoll = args.usecoll
```
murny: You could just pass the args into the method below instead of assigning it to a variable? Considering nothing is using this variable again.

PR #1358, opened 2017-03-03
Upgrade rubyzip to address vuln report

mbarnett:
@pbinkley @weiweishi All tests stayed passing and smoke-testing the application seemed fine. We should cut a new release and get this onto ERA test to verify.


murny:
:+1: :shipit:

PR #1360, opened 2017-03-20
1327 batch ingest update

### LINKED ISSUE #1327: batch:ingest_csv is under used and broken
Either fix this task or remove it.

weiweishi:
This is to fix #1327, and to allow us to start batch ingests again for ERA per product owner's request.


cwant:
There is some breakage due to this commit, in particular running:

bundle exec rake migration:era_collection_community['lib/tasks/migration/test-metadata/community']

This now fails because MigrationConstants isn't defined in several of the rake files (it was renamed to CommonConstants).

cwant:
New PR to fix the unresolved MigrationConstants references is here:

https://github.com/ualbertalib/HydraNorth/pull/1364

PR #1361, opened 2017-03-31
update gemfile

mbarnett:


Gemfile
```diff
@@ -57,6 +57,7 @@ gem 'sitemap', git: 'https://github.com/ualbertalib/rails-sitemap.git'
 gem 'rest-client'
 
 gem 'noid', '~> 0.8'
+gem 'nokogiri', '~> 1.7.1'
```
murny: Just a quick comment about this is for locking nokogiri or something for security reasons? As we ourselves are not using nokogiri in this project that I know of?

murny: Like this:

https://github.com/ualbertalib/HydraNorth/blob/496dcc631f7efc4b028b6f3515b45207182ab90b/Gemfile#L21-L22

mbarnett:
@pbinkley we should cut a release of hydranorth when we have a chance to resolve this gemnasium alert

pbinkley:
OK - we can do that Monday, when Henry's back to deploy to test

PR #1368, opened 2017-04-28
Install pry-rails gem.

cwant:
It's a nice package, but makes pry the default console.
To disable pry and use the old console do: ```DISABLE_PRY_RAILS=1 rails c```.


Gemfile
```diff
@@ -84,6 +84,7 @@ group :development, :test do
   gem "pry"
```
murny: Can probably remove 'pry' gem if your bringing in `pry-rails` as it supersedes it.

Also is anyone using pry-remote or pry-byebug?

cwant: Can remove pry from the Gemfile (according to Gemfile.lock, pry is a dependency of pry-rails, so it will get loaded in automatically anyways).

I have never used pry-remote or pry-byebug (to the best of my knowledge). "git blame" says that @mbarnett added them.

mbarnett: pry-byebug was to make it play nicely with byebug (a lot of the community seems big on byebug. Tricia uses it pretty heavily).

pry-remote is/was for connecting a pry-repl to ruby processes running under passenger. Useful if you were using the Vagrant dev environment, probably less useful now.

cwant: I think Tricia uses a vagrant dev environment, so I should keep pry-remote in for now.

mbarnett: OTOH, the odds that that environment still works by the time she's back are essentially nil.

cwant: Fine, will remove and push. (It can be added back if needed.)

cwant: Pushed.

PR #1374, opened 2017-06-19
Basic setup of Rollbar

murny:
https://github.com/ualbertalib/HydraNorth/issues/1375

Adds rollbar to Hydranorth. 

Add rollbar token to environment variable or secrets.yml in production to enable.


config/initializers/rollbar.rb
```diff
@@ -0,0 +1,53 @@
...
+  config.access_token = Rails.application.secrets.rollbar_token
+
+  # Without configuration, Rollbar is enabled in all environments.
+  # To disable in specific environments, set config.enabled=false.
+  config.enabled = Rails.env.production? && config.access_token.present?
+
+  # By default, Rollbar will try to call the `current_user` controller method
+  # to fetch the logged-in user object, and then call that object's `id`,
+  # `username`, and `email` methods to fetch those properties. To customize:
+  config.person_method = nil
```
murny: Couldn't find any documentation on how to disable this. 

But seems like they do an if check on this method and then proceed with setting up the personal data, so setting this to nil should prevent this:
https://github.com/rollbar/rollbar-gem/blob/d3ca4fbeceee7621cbe8964ad1e160d247c84afe/lib/rollbar/plugins/rails/controller_methods.rb#L9

cwant: I would wonder if this would be useful for debugging, e.g., does the issue affect all users or one specific user. Would it be worth obfuscating the user's identity by only exposing the id attribute, rather than turning the feature off? E.g.,

config.person_id_method = "id"
config.person_username_method = "id"
config.person_email_method = "id"

murny: Oh forsure good point and totally agree. I think this feature could be very useful. I like this suggestion to see if we can just turn off the email/username methods but keep the ID. I'll give an update with what I find out

murny: Updated this with what you suggested after talking to weiwei about it. Thanks!

cwant: One thought I had over lunch hour: since our view into the production environment is limited, we may not be able to map the id field to individual user records (at least not directly, without consulting a sys admin first).

murny: For sure, we won't be able to link the ID to an actual user without access to the database (which only sys admins are capable of). But at least we are capturing it and getting the general stats of "hey user 12345 is running into this same error every time they log in? What gives?" Which I am sure at this point if we want to investigate, the sys admins should have no problems with helping us with this?

Also it does seem like "admin" users have access to the users index page at least? https://github.com/ualbertalib/HydraNorth/blob/master/app/controllers/users_controller.rb#L9
Which probably has this information available of what ID corresponds to what user. Maybe this is motivation to do "admin" roles better in ERA?

mbarnett: Out of an abundance of paranoia/cutting off any possible source of objection from higher up, let's take it a step further and _hash_ the id, and just send the hash. We'll still be able to tell if an issue is affecting one user or many, because different users will have different hashes and any one user will work out to the same hash, but a theoretically attacker can't use the Rollbar data to even infer user ids or relative age of accounts off of sequential id information

mbarnett: vis a vis Chris' point:

> One thought I had over lunch hour: since our view into the production environment is limited, we may not be able to map the id field to individual user records

I think we should actively try to NOT be able to do this, in general, for a lot of reasons.

murny: I can get behind this. This will remove any source of objection like you said. I don't see much benefit of tracking down the actual user. The exception stack trace should give plenty of information itself. And at the end of the day, this already gives 1000% more information then we currently have right now.

But implementing this hashing will be more work for this, as we need a custom method on the user model for the hashing. 

def id_as_hash
   Digest::MD5.hexdigest(id)
end
```
 config.person_id_method = 'id_as_hash'
 config.person_username_method = 'id_as_hash'
 config.person_email_method = 'id_as_hash'
```

mbarnett: I'd do something like  

```ruby
def id_as_hash
   Digest::Sha2.hexdigest("#{some_secret_from_deployment_confg}_#{id})
end
```

The purpose of the secret is to act as a bit of salt, otherwise you can precompute a table by feeding incremental integers into the hash algorithm really easily and infer things about our database without access to them. It's paranoid but makes reversing more trouble than it's really worth.

murny: Good call :100:, I'll use `secret_key_base`

murny: Has been updated as discussed. ID is now being masked and hashed.

PR #1377, opened 2017-07-07
adding era-av button to era hompage

nnunn:


nnunn:
need to make one more change.

PR #1379, opened 2017-07-07
Docker and Docker Compose

murny:
https://github.com/ualbertalib/di_docker_hydranorth/issues/11

Docker using Docker Compose for Hydranorth


config/initializers/resque_config.rb
```diff
@@ -1,5 +1,4 @@
-config = YAML::load(ERB.new(IO.read(File.join(Rails.root, 'config', 'redis.yml'))).result)[Rails.env].with_indifferent_access
-Resque.redis = Redis.new(host: config[:host], port: config[:port], thread_safe: true)
+Resque.redis = $redis
```
murny: Why go and grab the information again and setup yet another redis connection? You already setup redis in the redis initializer...

config/initializers/redis_config.rb
```diff
@@ -1,12 +1,14 @@
...
 
-# use config/redis.yml to load settings
-redis_config_path = "#{rails_root}/config/redis.yml"
-redis_config = YAML.load(ERB.new(IO.read(redis_config_path)).result)
+# We should not even have a redis.yml file...
+# Just use REDIS_URL or fallback to the defaults which redis does automagically
+# Have to undo this stupid if we are using a ENV VAR now...
 
 # initialize redis connection
-Resque.redis = redis_config[rails_env]
```
murny: So much silliness here, fix alot...still room for improvement.

solr_conf/conf/solrconfig.xml
```diff
@@ -191,20 +191,10 @@
     </autoSoftCommit>
 
   </updateHandler>
-  <requestHandler name="/admin/" class="org.apache.solr.handler.admin.AdminHandlers" />
```
murny: This is depreacted and removed in higher versions of solr?

solr_conf/conf/schema.xml
```diff
@@ -301,33 +301,33 @@
...
-    
+
     <!-- An alternative geospatial field type new to Solr 4.  It supports multiValued and polygon shapes.
       For more information about this and other Spatial fields new to Solr 4, see:
       http://wiki.apache.org/solr/SolrAdaptersForLuceneSpatial4
     -->
     <fieldType name="location_rpt" class="solr.SpatialRecursivePrefixTreeFieldType"
-      geo="true" distErrPct="0.025" maxDistErr="0.000009" units="degrees" />
-    
+      geo="true" distErrPct="0.025" maxDistErr="0.000009" distanceUnits="degrees" />
```
murny: units is deprecated and removed in higher versions of solr?

solr_conf/conf/solrconfig.xml
```diff
@@ -11,10 +11,10 @@
   <directoryFactory name="DirectoryFactory" class="${solr.directoryFactory:solr.StandardDirectoryFactory}"/>
 
   <!-- solr lib dirs -->
```
murny: I dont think anything is actually using these solr_conf files? Considering we use jetty for development which has it's own solr configs. Who knows what we do in production?

spec-views/after.rb
```diff
@@ -1,14 +0,0 @@
-require "yaml"
```
murny: Removed spec-views folder....no idea what there for or how to even run them....did we write our own custom integration framework or something?

spec/controllers/my/files_controller_spec.rb
```diff
@@ -21,19 +21,24 @@
...
-        # does show normal files
-        expect(document_list_ids).to include(@my_file.id)
-        # doesn't show files shared with other users
-        expect(document_list_ids).to_not include(@unshared_file.id)
-      end
+    it "should respond with success and shows the correct documents" do
+      get :index
+      expect(response).to be_successful
+
+      document_list_ids = assigns[:response].docs.map(&:id)
```
murny: Mostly just style changes here...I did change this variable defination however. 

@document_list variable is not used anywhere in the view, everything is driven off the @response.docs . So let's test the thing that actually matters?

spec/features/collection/users_shouldnt_see_edit_collection_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
```
murny: Renamed this file....no idea what `1008_<filename>_spec.rb` means?

spec/features/generic_file_new_form_spec.rb
```diff
@@ -1,88 +1,74 @@
 require 'spec_helper'
```
murny: This is a really bad test. First its a form test, this should be testing against form class. But there isn't a generic_file_new_form class, so we are doing a integration test instead? Let's move this in the right folder then....

spec/features/generic_file_new_form_spec.rb
```diff
@@ -1,88 +1,74 @@
...
-      Capybara.default_max_wait_time = 30
-    end
+  describe 'new item fields', js: true do
+    # This is a very poor test... Sufia being stupid I am assuming?
+    # we are checking if we have a select tag that is not multiple but still uses an array???
     it "should not allow multiple resource_type selections, but assign to an array" do
-      sign_in user
-      visit '/files/new'
-      within ("#local #fileupload") do
+      skip 'Test fails if stack is being slow, very flaky, disable this as not really testing anything of value'
```
murny: I skipped these two test as they were flip flopping depending on how fast the page loaded. Very very britle.

Ultimately these tests aren't testing much, they literally depositing an item and then checking if we have the right form elements on the next page...(and they take up 10% of the entire test suite time, these tests take almost 2 minutes to run and total test suite time is 12-14 minutes...so yeah)

spec/features/generic_file_new_form_spec.rb
```diff
@@ -1,88 +1,74 @@
...
+
+      within("form#fileupload") do
```
murny: I thought these scopes were a bit weird, however I learned we actually have two forms on the deposit page, with much of the same ids and various other things :confused:

spec/features/generic_file_new_form_spec.rb
```diff
@@ -1,88 +1,74 @@
 require 'spec_helper'
-describe 'generic file new', :type => :feature do
-  let(:user) { FactoryGirl.create :user }
-  let!(:community1) do
```
murny: Moved all this data setup to where its actually needed. Only one test out of the 4 was using them.

spec/features/generic_file_new_form_spec.rb
```diff
@@ -1,88 +1,74 @@
...
-  after :all do
-    cleanup_jetty
+  before do
+    sign_in user
+    visit '/files/new'
   end
 
   describe 'unfold agreement div', js: true do
-    before do
-      visit '/'
```
murny: this is silly, why go to the homepage and then go to the deposit form...lets just go to the deposit form

spec/features/generic_file_new_form_spec.rb
```diff
@@ -1,88 +1,74 @@
...
+    after do
+      cleanup_jetty
     end
-    it "Title and creator is blank" do
-      sign_in user
-      visit '/files/new'
-      within ("#local") do
+
+    it "sets up fields properly" do
+      skip 'Test fails if stack is being slow, very flaky, disable this as not really testing anything of value'
```
murny: skip for same reason above

spec/jobs/create_derivatives_job_spec.rb
```diff
@@ -19,15 +19,13 @@
...
       @generic_file.save!
     end
-    context 'with a video (.avi) file', unless: $in_travis do
+    context 'with a video (.avi) file' do
       let(:mime_type) { 'video/avi' }
       let(:file_name) { 'countdown.avi' }
 
-      it 'lacks a thumbnail' do
+      it 'generates a thumbnail on job run' do
+        skip 'hydra-derivatives require proprietary custom codecs that are not included in ffmpeg, skip this test'
```
murny: skipped this test as it has a crazy infrasture requirement...you need a very specially built ffmpeg for this to even work...

spec/jobs/create_derivatives_job_spec.rb
```diff
@@ -19,15 +19,13 @@
...
         expect(@generic_file.thumbnail).not_to have_content
-      end
-
-      it 'generates a thumbnail on job run', :integration => true do
```
murny: no idea why a job test needs to be an integration test?

spec/spec_helper.rb
```diff
@@ -29,8 +29,6 @@
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 
-$in_travis = !ENV['TRAVIS'].nil? && ENV['TRAVIS'] == 'true'
```
murny: If your skipping tests for travisCI, that should be a flag that something isn't right with the test itself.

spec/support/vcr.rb
```diff
@@ -11,4 +11,11 @@
 
   # Only want VCR to intercept requests to external URLs.
   config.ignore_localhost = true
+
+  # Ignore any solr/fedora requests
+  config.ignore_request do |request|
+    if ENV['FCREPO_URL'] ||  ENV['SOLR_URL']
+      request.uri.include?(ENV['FCREPO_URL']) || request.uri.include?(ENV['SOLR_URL'])
```
murny: Now that fedora/solr are not on the same machine (when using docker), we need to tell VCR to ignore any traffic to fedora/solr

config/blacklight.yml
```diff
@@ -1,10 +1,10 @@
 development:
   adapter: solr
   url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/development" %>
-test: &test
```
murny: test variable never being used anywhere...removed.

docker-compose.yml
```diff
@@ -0,0 +1,85 @@
...
+services:
+  mysql:
+    image: mysql
+    environment:
+      MYSQL_ROOT_PASSWORD: mysecretpassword
+    volumes:
+      - mysql:/var/lib/mysql
+
+  solr:
+    image: murny/docker-solr-legacy:5.4
```
weiweishi: Can we put this in ualbertalib repository?

docker-compose.yml
```diff
@@ -0,0 +1,85 @@
...
+    # ports:
+    #   - "8983:8983"
+    volumes:
+      - solr:/opt/solr/server/solr/mycores
+      - ./solr_conf/conf:/config
+      - ./solr_conf/scripts/solr-precreate-hydranorth.sh:/docker-entrypoint-initdb.d/solr-precreate-hydranorth.sh
+
+
+  fcrepo:
+    image: murny/docker-fcrepo4:4.2
```
weiweishi: same here, moving it to ualbertalib repo?

solr_conf/conf/solrconfig.xml
weiweishi: We do use the solrconfig.xml in production. And in the past we work with sysadmins to test these solrconfig files during solr upgrade in test/staging environment.

solr_conf/conf/solrconfig.xml
weiweishi: do you know which higher versions? (5 or 6?)

spec-views/after.rb
weiweishi: we did at the beginning when we had limited knowledge about rspec, but it was not really used. They should've been removed a long time ago.

spec/controllers/my/files_controller_spec.rb
weiweishi: :+1:

spec/features/generic_file_new_form_spec.rb
weiweishi: Any other way we could validate the fields available on each deposit form, if not in the integration tests? Could we do them as a form test?

weiweishi (changes requested):

spec/features/generic_file_new_form_spec.rb
murny: Absolutey, you dont need an integration test to test if your form is setup right. Theses are very poorly written tests. It's like testing
` a = 2; assert a == 2;`
or in our case, `form is X, assert form is X`

One test is for testing if resource_types field is setup properly, we could clean up that view logic and move it into a helper or something and test the helper. Or test that  `Sufia.config.resource_types` contains the resource types you want. Plus we hacking some stuff in the view to make it work with sufia. Cause sufia expects resource_types to be multivalue, but we are changing the form to be single value. This logic SHOULD NOT be in the view. Never trust the user. A smart user like myself, can go edit the html/form or send a curl, and now I can assign multiple resource types to what is apparently a single value field. What problems will this cause?

config/solr.yml
```diff
@@ -1,7 +1,7 @@
 # This is a sample config file that points to a solr server for each environment
 development:
-  url: http://localhost:8983/solr/development
+  url: <%= ENV['SOLR_URL'] || 'http://localhost:8983/solr/development' %>
 test:
-  url: <%= "http://127.0.0.1:#{ENV['TEST_JETTY_PORT'] || 8983}/solr/test" %>
+  url: <%= ENV['SOLR_TEST_URL'] || 'http://localhost:8983/solr/test' %>
```
murny: Yes exactly. Both have same hosts/ports just different urls. 
`SOLR_URL` = `'http://solr:8983/solr/development`
and `SOLR_TEST_URL` = `'http://solr:8983/solr/test`

I suppose we could have just one as `SOLR_URL` = `http://solr:8983/solr/` and append the development/test to each in this solr.yml.

solr_conf/conf/solrconfig.xml
murny: Are you sure about that? Don't we shared solrs with other apps (does that mean we sharing configs)?  Are you sure ansible not overwritting them? etc? Sure be nice to have some documentation around these things. Couldn't find anything relating to this in `ansible-config`. And ansible-dev/local dev use jetty's solr config.

solr_conf/conf/solrconfig.xml
murny: I believe solr is quite good in this regards. This stuff is deprecated in version 5. And removed in version 6. It is not needed anymore.

solr_conf/conf/solrconfig.xml
mbarnett: Some part of it is being used. I added https://github.com/ualbertalib/HydraNorth/blob/master/solr_conf/conf/schema.xml#L356 to work with changes I made to solrizer, and it works in dev/test, so jetty's picking that up in those environments.

In Production I think we're using Solr Cloud (?? documentation would once again be a huge benefit for us all here) but there must be some way these custom Hydra schemas get picked up by it...

mbarnett: eh, nevermind, I'm thinking of the schema changes in HN2. I'd love to see some docs related to how all this solr config gets sorted out in different environments.

murny: No your correct, it seems like jettywrapper does copy solr/fedora configs out of certain directories: 
https://github.com/samvera/jettywrapper/blob/master/tasks/jettywrapper.rake

```

murny: To add to this, by default Jetty does not use your solr conf stuff! It will simply download the archive from `Downloading jetty at https://github.com/projecthydra/hydra-jetty/archive/v8.2.1.zip` which has its own solr/fedora configs (for solr its here https://github.com/samvera/hydra-jetty/tree/v8.2.1/solr).

If you want to use your own you can use the jetty:config (which i have never done before until today).

PR #1380, opened 2017-07-18
add staging environment to hydranorth for rollbar implementation

weiweishi:
Added proper staging environment to facilitate testing of the rollbar configuration. 

Additional ansible playbook changes are required to set proper environment variables for the staging environment.


config/database.yml
```diff
@@ -1,31 +1,33 @@
...
-
-# Warning: The database defined as "test" will be erased and
-# re-generated from your development database when you run "rake".
-# Do not set this db to the same as development or production.
 test:
   adapter: mysql2
   database: hydranorthtest
   username: root
-  password:
+  password:
```
murny: Your editor is leaving some extra spaces on the password keys?

config/database.yml
```diff
@@ -1,31 +1,33 @@
...
-  password:
+  password: 
   pool: 5
   timeout: 5000
-
+staging:
+  adapter: mysql2
+  database: hydranorthstaging
+  username: hydranorth
+  host: mysql-test
```
murny: isn't most of this config from ansible anyways? But nice to bring over thats for sure

config/redis.yml
```diff
@@ -4,6 +4,10 @@ development:
 test:
   host: localhost
   port: 6379
+staging:
+  host: ENV['REDIS_URL']
```
murny: I'd leave this localhost?

REDIS_URL is a bit more then the host, it's actually this enitre config. You can provide it with a url key though if you want to use the env var:

You can also specify connection options as a redis:// URL:

`redis = Redis.new(:url => "redis://:p4ssw0rd@10.0.1.1:6380/15")`

see here: https://github.com/redis/redis-rb#getting-started

config/secrets.yml
```diff
@@ -24,6 +24,13 @@ test:
   ezid_password: <%= ENV['EZID_PASSWORD'] %>
   preservation_queue_name: <%= ENV['PMPY_QUEUE_NAME'] || 'test:pmpy_queue' %>
   rollbar_token: ''
+staging:
+  secret_key_base: 450fc6b88e8a27fe9ca23621599762b9faab14421121131d98f6f4de737cd8a74266fd7e2c0f7316a2388066722ea01e354bb9930f4ea223bd7bd02cf11e65c3
```
murny: should be the env var: `secret_key_base: <%= ENV['SECRET_KEY_BASE'] %>`

You want to `rake secret` this to generate this token. Otherwise having this public opens a whole can of security worms

config/database.yml
weiweishi: I like this approach. Will modify the file accordingly.

config/redis.yml
weiweishi: Great. thanks! I'd say we use the env vars whenever possible? I'll modify the file accordingly.

config/secrets.yml
weiweishi: yes. should've changed it.

config/database.yml
```diff
@@ -1,31 +1,53 @@
...
 # Do not set this db to the same as development or production.
 test:
-  adapter: mysql2
+  <<: *default
   database: hydranorthtest
-  username: root
-  password:
-  pool: 5
-  timeout: 5000
+  url: <%= ENV['DATABASE_URL'] %>
```
murny: something got mangled here? Needs a staging namespace above this. After this is fixed, ill reapprove

murny: You can remove this one. Just needs the staging/production.

Assuming you copied this out of jupiter (this is here in the test namespace for a different reason https://github.com/ualbertalib/jupiter/blob/master/config/database.yml#L31-L36 as rails 5 will provision both dev/test DBs by default)

weiweishi: Ah, okay, makes sense. hope third time's the charm.

PR #1382, opened 2017-08-02
Staging env

weiweishi:
Update configuration files and secrets.yml, to use secrets.yml file as the single source of variables needed for the application configuration in different environments.


config/secrets.yml
```diff
@@ -17,20 +17,65 @@ development:
   ezid_password: <%= ENV['EZID_PASSWORD'] %>
```
murny: You could create a yaml var for default here and import it under all the environment namespaces too if you want to reduce some code here. Considering staging/prod are exactly the same (or dev/test are very similar)

config/environments/staging.rb
```diff
@@ -86,4 +87,4 @@
 end
 
 # Required when using Rails.application.routes.url_helpers from outside the request/response life cycle (models, jobs, lib)
-Rails.application.routes.default_url_options = { host: 'era-test.library.ualberta.ca', port: 443, protocol: 'https' }
+Rails.application.routes.default_url_options = { host: Rails.application.secrets.default_url_host, port: Rails.application.secrets.default_url_port, protocol: 'https' }
```
murny: Not sure you would ever change the default url port here (from 443), but i guess this doesnt hurt

config/environments/production.rb
```diff
@@ -64,13 +64,13 @@
...
-  config.action_mailer.default_url_options = { :host => 'localhost' }
+  config.action_mailer.default_url_options = { :host => Rails.application.secrets.devise_mailer_url }
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.smtp_settings = { :address => "localhost", :port => 25 }
-  config.action_mailer.default :from => 'hydranorth@mailinator.com'
+  config.action_mailer.default :from => Rails.application.secrets.devise_email
 
   #contact for config
-  config.contact_email = 'hydranorth@mailinator.com'
+  config.contact_email = Rails.application.secrets.contact_email
```
murny: Have a feeling this contact_email is never being used.

Looks like it was introduced here and nothing references it:
https://github.com/ualbertalib/HydraNorth/commit/de119a12d76ce0d2052e4986cce07ae4928b4b4f#diff-a2f09e1db7a47eb67c5849478a97f3d7

weiweishi: yeah, it is referenced in here: https://github.com/ualbertalib/HydraNorth/blob/c949a558222104335a098bcd495aed726a6c25da/app/models/contact_form.rb but I don't think we use Contact Us form.

config/environments/staging.rb
weiweishi: good point. I was capturing what's been overwritten in the Ansible playbook, but I think you are right, no need to have a port that probably will never change into a variable.

config/secrets.yml
weiweishi: Changed accordingly.

config/environments/production.rb
murny: well thats the `Sufia.config.contact_email` which should be a different config (its from the sufia intializer)

and yeah we arent using the contact form, its currently just a static page with some email addresses on it.

So ¯\_(ツ)_/¯

PR #1384, opened 2017-08-04
properly render the values from secrets.yml in config files

weiweishi:
Apparently <%= %> is needed in the config/*.yml files to properly render the values from secrets.yml.


weiweishi:
@murny I hope this will be the last one of this series. It's now tested and working fine on era-test.

murny:
Hopefully sufia is running those files through ERB preprocessor which I think they are? So if you get any that don't work that's why.

weiweishi:
Okay, makes sense. Thank you!

PR #1386, opened 2017-08-16
additional resque config change to initialize with proper redis setup

weiweishi:
@murny This config works on era-test servers, but wondering if there are other better ways to handle the resque config so we don't have to initialize two redis instances?


murny:
I would steal the changes from my docker PR #1379 ....at this point...why have a redis.yml and why do stupid for both redis/resque?

https://github.com/ualbertalib/HydraNorth/pull/1379/files/7d61749883dda06f985fd0bc00c8681fad0245f7#diff-188503b6c310fb4f3ac5ed9eb9f81a49

https://github.com/ualbertalib/HydraNorth/pull/1379/files/7d61749883dda06f985fd0bc00c8681fad0245f7#diff-47dbc8225025d7a1c65fb7f3fba5ee7e

You can simply just set the url from the env/secrets.yml and call it a day. Then do the resque fix i did in my PR

weiweishi:
Exactly. Your change makes total sense to me. (Sorry I had forgotten about checking there first). Will make changes accordingly.

weiweishi:
@murny made changes accordingly and tested.

config/initializers/redis_config.rb
```diff
@@ -1,12 +1,9 @@
...
 # initialize redis connection
-Resque.redis = redis_config[rails_env]
-
-# pull out the parsed redis driver from resque
-$redis = Resque.redis.instance_eval{@redis}
+if ENV['REDIS_URL']
+  $redis = Redis.new
+else
+  # use config/secrets.yml to load settings
+  $redis = Redis.new(:url=> Rails.application.secrets.redis_url)
```
murny: Would just collapse these two into one like so: 

```ruby
Redis.current =  Redis.new(url: ENV['REDIS_URL'] || Rails.application.secrets.redis_url)
```

Would use `Redis.current` over a global ruby variable. Also use ruby 1.9 hash syntax instead of hash rockets ( `=>`)

weiweishi: Thanks. And learnt something new re: [ruby hash style](https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md#hashes). Also, I'm curious, what is the loading order of the initializers .rb files when the app is initialized? alphebatical order? i.e. Redis.current (or $redis) is initialized before loading resque_config.

murny: its alphabetical order yes, so when you open the folder, the first intializer runs first. If you need to change or make sure of a certain order, you can prefix your intiailziers with numbers.

Heres an example of this in the wild: https://github.com/discourse/discourse/tree/master/config/initializers

So we okay here since redis comes before resque.

weiweishi: Thank you!

PR #1410, opened 2017-08-24
updated robot.txt to avoid crawling the item stats page

weiweishi:
All the error message related to #1399 are caused by bots trying to crawl the item stats page. The stats page does send an API request to Google Analytics. By blocking the bots from crawling this page, we will fix the majority occurrences of this issue, and can see if anything else could be causing it.


weiweishi:
Thanks @cwant

PR #1517, opened 2017-09-08
remove unnecessary error handling in pushing to preservation queue

weiweishi:
To address all the "Can't Preserve..." error messages in Rollbar


lib/hydranorth/preservation_queue.rb
```diff
@@ -11,7 +11,6 @@ def queue
   def preserve(noid)
     queue.with do |conn|
       status = conn.zadd QUEUE_NAME, Time.now.to_f, noid
-      raise PreservationError unless status == true
```
murny: Since nothing is using PreservationError now lets remove this as well?

On LN#3: https://github.com/ualbertalib/HydraNorth/blob/b2f3ee0a9b38760d96f55571162d1fc991a80d0c/lib/hydranorth/preservation_queue.rb#L3

PR #1534, opened 2017-09-25
Upgrade nokogiri to post-usn-3424-1

mbarnett:
Upgrade version of Nokogiri to deal with Gemnasium complaints

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,8 +58,8 @@ gem 'rest-client'
 
 gem 'noid', '~> 0.8'
 
-# pin this to post-USN-3271-1
-gem 'nokogiri', '~> 1.7.2'
+# pin this to post-USN-3424-1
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -58,8 +58,8 @@ gem 'rest-client'
 
 gem 'noid', '~> 0.8'
 
-# pin this to post-USN-3271-1
-gem 'nokogiri', '~> 1.7.2'
+# pin this to post-USN-3424-1
+gem 'nokogiri', '~> 1.8.1'
```
mbarnett: Somehow ironic that XML is our number one source of security-related upgrades

murny: XML is pretty crazy though: https://blog.pragmatists.com/xml-be-cautious-69a981fdc56a

PR #1798, opened 2018-12-06
Hyperlink DOIs to preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). It may be a bit ironic that they would change the URL of their service, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org). Because the old `dx` subdomain is being redirected, there is no urgent need to do anything.

However, I'd hereby like to suggest to follow the new recommendation by applying this update to all static DOI links and the code that generates new DOI links.

Cheers!


katrinleinweber:
I don't think the [Travis errors](https://travis-ci.org/ualbertalib/HydraNorth/builds/464182858#L623) have anything to do with these changes.

pgwillia:
Thank you for your input! It's a good suggestion. 

I wanted to let you know we no longer use this repository (it was replaced by https://github.com/ualbertalib/jupiter), so the code you've found is not being maintained or used at our institution.

katrinleinweber:
OK, thanks for explaining. I'd argue that somebody else may still be using your code, and might have benefitted from this being merged. Either way, you could [archive this repo](https://help.github.com/articles/archiving-repositories/) then ;-)

